PR #2, opened 2017-04-26
amendment

gleamy:
amendment

### DIFF
--- a/src/Clients/MTConnectClient.cs
+++ b/src/Clients/MTConnectClient.cs
@@ -203,19 +203,34 @@ namespace MTConnect.Clients
                     int itemCount = -1;
                     if (doc.DeviceStreams.Count > 0)
                     {
-                        var deviceStream = doc.DeviceStreams.Find(o => o.Name == DeviceName);
-                        if (deviceStream != null && deviceStream.DataItems != null)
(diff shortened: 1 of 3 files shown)


PatrickRitchie:
Thanks for taking the time to submit a Pull Request! I am however a little confused as to what you are wanting to accomplish with the changes. Can you add a description?

It looks like you are wanting to add the ability to read from multiple Devices using the same MTConnectClient class. This class was intended to be used for a single device although if this is something you and/or others are wanting, we may need to change a few other things as well to get the raised events to work correctly. Typically multiple devices are read by a Probe request first, then for each Device found, a MTConnectClient is started for each.

gleamy:
Thanks for your reply

I just want to when we are not sure to collect the device,MTConnectClient can return all of the data for us.

In addition, MTConnectClient did not say it is only for a single device

PatrickRitchie:
Sorry for the delay. We are looking over what it will take to make this work correctly. The main issue is how to handle the AssetChanged data item which will now have to be stored for each device. Shouldn't be too difficult but does change a few things. I'll make sure to include this along with the changes you made to the Url.

PatrickRitchie:
This is addressed with v3.0.0

PR #8, opened 2019-07-15
Added Accept header to all requests

Jandella:
While I was testing the library on address http://mtconnect.mazakcorp.com:5612, I received a null document.
Upon debugging, I found the response was in json format instead xml.

Adding the accept header to receive a xml response resolved the issue.


Jandella:
No prob ;)

PR #13, opened 2021-01-13
Added handling of TaskCanceledException for all occurrences of HttpClient

smedsbo:
If this exception is unhandled the thread will terminate if a timeout occur in the httpClient


PatrickRitchie:
This has been addressed with v3.0.0

PR #49, opened 2023-05-19
TopicPrefix propagation to mqtt relay agent

M2rtViljaste:
Add topic prefix propagation from agent config file to mqtt relay agent


PatrickRitchie:
Thanks for update! Looks like everything will work fine. I did debate whether or not the TopicPrefix property should be read only but I think I like the way you have it where it is only set in the constructor through the configuration.

Thanks,
-Patrick

Issue #138, opened 2026-04-24
`JSON-CPPAGENT-MQTT` Probe emits `"name": ""` for DataItems whose `Name` is set to empty string (XML formatter correctly omits)


ottobolyos:
## Summary

Under the `JSON-CPPAGENT-MQTT` document format, Probe DataItems whose caller did not explicitly set `Name` are emitted with `"name": ""` (empty-string placeholder) rather than the `name` key being omitted. The sibling `MTConnect.NET-XML` formatter already handles this correctly — it omits the attribute when `Name` is null or empty. The JSON-cppagent copy path is unconditional and needs the same guard.

Downstream consumers reading Probe DataItems by name cannot distinguish "the DataItem has no name" from "the DataItem has name equal to empty string"; the empty-string serialisation is spec-ambiguous.

## Environment

- `MTConnect.NET-JSON-cppagent` + `MTConnect.NET-Applications-Agents` at **v6.9.0.2** (official Docker image `trakhound/mtconnect.net-agent:6.9.0`, published 2025-10-16 — binary-equivalent to the current NuGet release).
- Broker: `eclipse-mosquitto:2.0.22`.
- Comparison agent: `mtconnect/agent:latest` = cppagent v2.7.0.7 (for spec-conformant behaviour reference).
- Format ID: `JSON-CPPAGENT-MQTT`.

## Reproduction

Minimum rig — programmatic Device construction (the same surface `dime-connector` exposes via its YAML xpath construction, but reproducible from any consumer of the library):

Reproduced at HEAD (commit `3d6321ab` on master, 2026-04-24) via a minimal console program that constructs `JsonDataItem` instances directly — no consumer-specific plumbing, so the evidence is library-intrinsic:

```csharp
using MTConnect.Devices;
using MTConnect.Devices.DataItems;
using MTConnect.Devices.Json;

// Case 1 — typed DataItem, Name not set in initialiser
var c1 = new TemperatureDataItem { Id = "t1", Category = DataItemCategory.SAMPLE, Units = "CELSIUS" };
Console.WriteLine(new JsonDataItem(c1));

// Case 2 — Name explicitly set to empty string (the defect)
var c2 = new TemperatureDataItem { Id = "t2", Name = "", Category = DataItemCategory.SAMPLE, Units = "CELSIUS" };
Console.WriteLine(new JsonDataItem(c2));

// Case 3 — Name explicitly non-empty (the working case, for contrast)
[...]

Issue #137, opened 2026-04-23
`new Device()` ctor stamps demo-fixture defaults (`Id="4A1GF40513"`, `Name="dev"`); `new <T>Component()` ctors back-fill `Name` from lowercase type-name


ottobolyos:
## Summary

Two related library-quality observations against `MTConnect.NET-Common` v6.9.0.2. Neither is a direct MTConnect spec violation — the XSD requires `id` / `name` on the wire, and a caller-supplied value overwrites the ctor default — but both are surprises that have leaked into at least one third-party project on GitHub (search for `"4A1GF40513"` returns hits in unrelated MT.NET consumer repos that forgot to set `Id`).

This draft is filed separately from the companion issue #136 (`Device.Uuid` ctor leak) because the `Uuid` case has a concrete spec-contract violation to argue, whereas the two below are best-practice concerns for a library POCO constructor.

### Observation A — `new Device()` stamps demo-fixture values in `Id` and `Name`

`Device`'s default constructor assigns:

```
Id   = "4A1GF40513"      // exact value; shipped in the library's sample Devices.xml
Name = "dev"              // exact value; shipped in the library's sample Devices.xml
```

`"4A1GF40513"` and `"dev"` are the values used by the library's own sample fixture — they appear verbatim in the demo `Devices.xml` included with the library. A consumer constructing `new Device { Uuid = "F1" }` and shipping it without also setting `Id` and `Name` ends up broadcasting the library's fixture strings on the wire.

### Observation B — `new <T>Component()` ctors stamp lowercase type-name as `Name`

`LinearComponent`, `AxesComponent`, `SystemsComponent` (and, by inference, other registered Component subclasses) have ctors that back-fill `Name` from the lowercase type name:

```
new LinearComponent() .Name == "linear"
new AxesComponent()   .Name == "axes"
new SystemsComponent().Name == "systems"
```

`Id` and `Uuid` stay `null` on these — only `Name` leaks. A consumer who constructs `new LinearComponent { Id = "x" }` without setting `Name` ends up with a `Name = "linear"` that nothing in their own code produced. The leaked value is at least deterministic from the type (unlike the random `Uuid` case), but it is still library-authored content a consumer may not realise is present.

## Verified by reproduction — 2026-04-23

Two xUnit probes against the library assembly:

```csharp
// Probe 1 — Device
var d = new Device();
// observed: Id='4A1GF40513'  Name='dev'  (Uuid covered separately)

[...]

Issue #136, opened 2026-04-23
`new Device()` ctor auto-generates a random `Uuid` that re-rolls on every construction, silently violating the XSD "for its entire life" identity contract when a caller does not overwrite it


ottobolyos:
## Summary

`MTConnect.NET-Common`'s `Device` default constructor assigns `Uuid = Guid.NewGuid().ToString()` — **a freshly-generated random GUID on every construction**. The value is a public setter that a caller's object-initializer can overwrite, but there is no observable difference between "I explicitly set this" and "I relied on the default, and the library rolled a fresh GUID that nothing in my code generated". The leak is therefore hidden.

Consequence in production: every consumer that doesn't set `Uuid` explicitly registers a **different** Uuid on every connector restart, because the ctor re-rolls the GUID. Downstream stores (dashboards, historians, bridges) that key on `Device.Uuid` — the canonical identity MTConnect has defined since v1.0 (2008) — end up with **one row per connector restart** under the same `Device.Name`. Observed live at Tempco 2026-04-23: 4 rows for a single `Furnace1West` device, 2 for `Furnace2East`, accumulated over a handful of pre-merge restarts of our connector.

This draft is narrowly about the `Uuid` slot because it is the only one of the three ctor-leaked values whose spec contract is directly violated by random-per-construction generation. The related `Id` and `Name` fixture defaults (`Id = "4A1GF40513"`, `Name = "dev"`) are a separate library-quality concern tracked in a sibling draft.

## Verified by reproduction — 2026-04-23

Reproduced live with `MTConnect.NET-Common` as referenced by `MTConnect.NET-Applications-Agents v6.9.0.2` on `net9.0` Linux x64:

```csharp
var d = new Device();
// observed: Uuid='e6b40112-99a2-40d8-9bed-a5bbbdf16177'
// second construction: Uuid='b7c3e820-…' — different every time
```

Two sequential `new Device()` calls produce Devices with different `Uuid` values. Setting `Uuid` in the object initializer (`new Device { Uuid = "F1" }`) correctly produces `Uuid = "F1"` — the setter has precedence — but this requires the caller to remember to set every single `Device` they construct.

## Environment

- `MTConnect.NET-Common` via `MTConnect.NET-Applications-Agents v6.9.0.2` (current latest release, published 2025-10-16; same assembly ships in Docker image `trakhound/mtconnect.net-agent:6.9.0`).
[...]

Issue #135, opened 2026-04-21
`mqtt-relay` module publishes Agent Availability on a non-canonical `{prefix}/Probe/{agent-uuid}/Available` four-segment topic


ottobolyos:
## Summary

The `MTConnect.NET-AgentModule-MqttRelay` module publishes the Agent Device's Availability state on `{topicPrefix}/Probe/{agent-uuid}/Available` — a four-segment topic under the `Probe/#` wildcard — both as the MQTT Last-Will-Testament (LWT) payload and as a retained "agent is up" message on connect. The payload is a **raw unquoted string** (`AVAILABLE` / `UNAVAILABLE`), not a JSON Probe envelope. This collides with the canonical topic contract that every topic under `{prefix}/Probe/#` carries a JSON envelope parseable by any MTConnect document consumer — subscribers using a `Probe/#` wildcard receive the non-JSON payload and fail to parse.

## Environment

- `MTConnect.NET-Applications-Agents` + `MTConnect.NET-AgentModule-MqttRelay` at **v6.9.0** (official Docker image `trakhound/mtconnect.net-agent:6.9.0`, published 2025-10-16 — binary-equivalent to v6.9.0.2).
- Broker: `eclipse-mosquitto:2.0.22`.
- Module configured with `topicStructure: Document` (the default that encourages full-document publishes on Probe / Current / Sample / Asset).
- Comparison agent: `mtconnect/agent:latest` = cppagent v2.7.0.7 — which publishes only three-segment topics under `{prefix}/Probe/#`.

## Reproduction

Minimum rig — `mqtt-relay` module with `topicStructure: Document`:

```yaml
# applications/MTConnect-Agent/appsettings.yaml
modules:
  - mqtt-relay:
      server: localhost
      port: 1883
      documentFormat: JSON-CPPAGENT-MQTT
      topicPrefix: MTConnect/Document
      topicStructure: Document
```

Run the agent, subscribe to the full Probe wildcard, and observe the four-segment topic:

```sh
mosquitto_sub -h localhost -t 'MTConnect/Document/Probe/#' -v
```

### Observed — MT.NET

Two topic-patterns appear:

```text
MTConnect/Document/Probe/<agent-uuid>              { "Agent": { … } }       ← canonical 3-segment Probe
MTConnect/Document/Probe/<device-uuid>             { "Device": { … } }      ← canonical 3-segment Probe
[...]

PR #139, opened 2026-04-25
feat: add MTConnect Standard v2.6 + v2.7 support (#133)

### LINKED ISSUE #133: Add support for MTConnect Standard v2.6 and v2.7
## Summary

The MTConnect standard has published versions **v2.6** and **v2.7** (latest), whose normative information model, XMI, and generated XSDs are available in the official MTConnect repositories. `MTConnect.NET` currently advertises support "up to MTConnect v2.5" on both the README and the NuGet descriptions, and no release on the public release track ships models, enumerations, or encoders for the v2.6 or v2.7 delta.

This issue asks the maintainers to extend the library to serve v2.6 and v2.7 — structural classes, DataItem sub-structures, enumeration additions, and any attribute additions — so that an agent can truthfully advertise `Header.version = "2.7.0.0"` and `schemaVersion = "2.7"` while emitting the fields the v2.7 XSD requires.

## Verified — 2026-04-20

Confirmed by inspection of the library source ([`libraries/MTConnect.NET-Common/MTConnectVersions.cs`](https://github.com/TrakHound/MTConnect.NET/blob/master/libraries/MTConnect.NET-Common/MTConnectVersions.cs)):

```csharp
public static Version Max => Version25;
// ... Version10 through Version25 ...
// No Version26 or Version27 defined.
```

The library's own version table stops at v2.5. No v2.6 or v2.7 entries exist. `MTConnectVersions.Max` is pinned at `Version25`. Live test confirms: a v6.9.0 agent running with `JSON-CPPAGENT-MQTT` format serves only up to the v2.5 semantic surface, and the relevant cppagent comparison image (`mtconnect/agent:latest` = v2.7.0.7) ships full v2.7 support for the same capability.

## Current state

| Fact                                            | Value                                                                                            |
|-------------------------------------------------|--------------------------------------------------------------------------------------------------|
| Library latest release                          | `v6.9.0.2` (2025-10-16)                                                                          |
| Advertised spec ceiling                         | MTConnect v2.5 (per README + NuGet description)                                                  |
| v2.5 support landed in                          | `v6.7.0` (2025-06-22)                                                                            |
| v2.4 support landed in                          | `v6.5.0` (2024-10-21)                                                                            |
| v2.6 / v2.7 support                             | **Not available in any published release**                                                       |
| Recent release focus                            | MQTT relay / SHDR timestamp fixes; no version-uplift work                                        |
| `MTConnect.NET-JSON-cppagent` module disclaimer | "This Module is still under development and may be deprecated in the future" (per NuGet listing) |

## Normative sources for v2.6 + v2.7

### XMI / SysML information model (normative — XSDs are generated from this)

- **SysML model repository**: <https://github.com/mtconnect/mtconnect_sysml_model>
  - Root file: `MTConnectSysMLModel.xml`
  - Released tags include **Version 2.7 (February 2026)** and earlier — the releases page lists 12 total.
- **Transpiler** (useful for code generation): <https://github.com/mtconnect/MtconnectTranspiler> — transpiles the SysML model into other formats.
- **Model browser** (for human-readable navigation of the information model): referenced from the `mtconnect_sysml_model` README.

### XSDs (generated from the XMI — useful for validation + direct reference)

- **Schema repository**: <https://github.com/mtconnect/schema>
- **Merge PR adding v2.6 + v2.7 XSDs**: [`mtconnect/schema` PR #25](https://github.com/mtconnect/schema/pull/25) (merged 2026-04-13).
- **Mirrored XSD host**: <https://schemas.mtconnect.org/> (CDN-hosted; all files at `/schemas/`).

Direct-link index for the eight relevant files (four schemas × two versions):

| Version | Devices                                                                                              | Streams                                                                                              | Assets                                                                                             | Error                                                                                            |
|---------|------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| **2.7** | [`MTConnectDevices_2.7.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectDevices_2.7.xsd) | [`MTConnectStreams_2.7.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectStreams_2.7.xsd) | [`MTConnectAssets_2.7.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectAssets_2.7.xsd) | [`MTConnectError_2.7.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectError_2.7.xsd) |
| **2.6** | [`MTConnectDevices_2.6.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectDevices_2.6.xsd) | [`MTConnectStreams_2.6.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectStreams_2.6.xsd) | [`MTConnectAssets_2.6.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectAssets_2.6.xsd) | [`MTConnectError_2.6.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectError_2.6.xsd) |

Each version also ships `_1.0` variants for legacy parsers that do not support XSD 1.1 — same filenames with `_1.0` suffix.

### Prose standard

- MTConnect Standard download: <https://www.mtconnect.org/standard-download20181> — free registration, covers Parts 1–5 (Overview, Devices, Streams, Assets, Interfaces). From v2.0 onwards the XMI is the normative information model and the XSDs + prose are generated from it.

## What v2.6 + v2.7 add (at a glance)

The v2.6 / v2.7 XSDs extend the v2.5 baseline along several axes; the maintainers will want to consult the XMI + XSD diff directly for the complete enumeration, but the known categories are:

- **New structural classes and sub-components**, particularly in the Configuration and DataItem sub-structures introduced in v2.6.
- **New enumeration values** on existing enumerated types (e.g. DataItem types, condition qualifiers, asset categories, event-value enumerations).
- **New optional attributes** on existing component types.

Spec-level wire-shape constraints (Header attribute lists, substitution-group chains for `Component`/`CommonComponent`/`System`, `FloatSampleValueType` union, etc.) are **unchanged** across v2.5 → v2.6 → v2.7. This issue therefore does not depend on the fixes for wire-shape defects (see the related-issues list below); uplift and wire-shape fixes are orthogonal and can land in either order.

## Scope proposal

Three layers that would benefit from uplift, in decreasing order of user-facing impact:

### Layer 1 — typed models + enumerations

Add v2.6 + v2.7 entries to the library's typed model surface:

- New structural classes for components / sub-components added in v2.6 and v2.7.
- New enumeration values for every existing `XxxType` / `XxxEnum`.
- New optional attribute fields on existing types.
- Version-keyed compatibility so the same assembly can serve v2.5, v2.6, and v2.7 concurrently (operators select via `DefaultVersion`).

### Layer 2 — XSD-aware validation

The library's `InputValidationLevel` setting currently validates against v2.5. Extending validation to v2.6 and v2.7 would:

- Load the appropriate schemas per configured `DefaultVersion`.
- Surface v2.6 / v2.7-specific attribute additions to operators.
- Let the library reject inputs that are malformed under the selected spec release.

### Layer 3 — encoders + schemaVersion wiring

Wire `Header.version` + `schemaVersion` + the XML and JSON encoders to the configured MTConnect release so that a v2.7 agent actually emits v2.7 payloads. This ties together with two companion issues filed in the same batch:

- #127 — *`Header.version` reports library assembly version* — the slot should report the configured MTConnect release.
- #128 — *`MTConnectStreams.schemaVersion` / `MTConnectDevices.schemaVersion` hardcoded to `"2.0"`* — the default should be the library ceiling.

Fixing both companion issues **plus** landing v2.6 / v2.7 support would let an unconfigured agent cleanly advertise the latest MTConnect release without any operator intervention.

## Suggested approach

1. **Generate-first** is the lowest-risk path. The normative XMI at <https://github.com/mtconnect/mtconnect_sysml_model> is the single source of truth; regenerating typed models and enumeration tables from the XMI (possibly via [`MtconnectTranspiler`](https://github.com/mtconnect/MtconnectTranspiler) or an in-repo equivalent) would catch v2.6 / v2.7 additions systematically rather than hand-translating. It also sets up the library to track future spec releases without per-release manual labour.
2. **Slice by XSD**: Devices → Streams → Assets → Error is a natural PR ordering, matching the four schemas and the likely independent review cadence.
3. **Coverage test fixture**: a fixture that validates a round-tripped Probe / Current / Sample / Assets payload against the corresponding XSD for every supported version catches typed-model regressions as they land.
4. **Version-ceiling detection helper**: consumers benefit from a public API like `MTConnectVersion.Latest` that resolves to the library's compiled-in ceiling; combined with the "default to ceiling" fix in the companion `schemaVersion` issue, this lets operators pin to the latest transparently.

## Relationship to the wire-shape / structural defects filed in this batch

| Companion issue                                                                      | Independent of uplift? | Notes                                                                                                                                                                                                              |
|--------------------------------------------------------------------------------------|------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| *`Header.version` reports library assembly version*                                  | ✅                      | The slot reports the library assembly version today; fixing the slot is independent of which MTConnect version the library supports. The fix enables truthful reporting of `"2.7.0.0"` once support lands.        |
| *`schemaVersion` defaults to library floor (`2.0`)*                                  | ✅                      | Default should be "the library ceiling"; the ceiling value changes per uplift release.                                                                                                                             |
| *Probe + Assets Header missing `assetBufferSize` / `assetCount`*                     | ✅                      | Both fields have been required continuously since v1.2 and are unchanged across v2.5 → v2.7.                                                                                                                       |
| *Numeric Sample `value` emitted as JSON string*                                      | ✅                      | `FloatSampleValueType` definition is unchanged across every v2.x release.                                                                                                                                          |
| *`Agent.AddDevice()` — `<Systems>` organiser auto-nesting inconsistent per subtype* | ✅                      | Substitution-group rules unchanged. Any new `System` subtypes added in v2.6 / v2.7 will inherit whichever branch the current per-subtype logic places them in — fixing this alongside the uplift prevents that.   |

So: every companion wire-shape / structural defect stands independent of this uplift, and the uplift stands independent of those fixes. The cleanest compliance posture requires both tracks to complete.

## References

- Spec:
  - SysML / XMI information model: <https://github.com/mtconnect/mtconnect_sysml_model>
  - SysML transpiler: <https://github.com/mtconnect/MtconnectTranspiler>
  - Schema repo: <https://github.com/mtconnect/schema>
  - Schema CDN: <https://schemas.mtconnect.org/>
  - Prose standard: <https://www.mtconnect.org/standard-download20181>
- cppagent reference (for behaviour parity): <https://github.com/mtconnect/cppagent>
- Related library issues in this repo (filed in the same batch; cross-reference numbers will be added after the sibling issues are opened):
  - *`Header.version` reports library assembly version*
  - *`schemaVersion` defaults to library floor (`2.0`)*
  - *Probe + Assets Header missing `assetBufferSize` / `assetCount`*
  - *Numeric Sample `value` emitted as JSON string*
  - *`Agent.AddDevice()` — `<Systems>` organiser auto-nesting inconsistent per `System` subtype*

ottobolyos:
## Summary

Extends MTConnect.NET to cover MTConnect Standard v2.6 and v2.7 (closes #133). v2.6 + v2.7 type-system + Configuration polymorphism are complete, wire-format end-to-end, and round-trip-tested; the L1-L5 compliance test harness layer is scaffolded with all-version XSDs imported.

### v2.7 type-system + Configuration polymorphism

- Add `MTConnectVersions.Version26` and `Version27`; advance `Max` to `Version27`.
- Regenerate Common DataItems / Components / Configurations + JSON-cppagent formatters from `mtconnect/mtconnect_sysml_model` tag `v2.6` (SHA `08185447bf86…`) and `v2.7` (SHA `25796ac591bb…`).
- New types in v2.6: `AssetAddedDataItem`, `AssociatedAssetIdDataItem`, `CuttingTorchComponent`, `ElectrodeComponent`. New enum value `MediaType.QIF_MBD`.
- New types in v2.7: `BindingState`, `Depth`, `FixtureAssetId`, `SwingAngle/Diameter/Radius`, `TaskAssetId`, `WaterHardness` DataItems; `PinTool`, `ToolHolder` Components; `Axis` / `Origin` / `Rotation` / `Scale` / `Translation` configuration primitives (each with `Abstract*` base + concrete + `*DataSet` data-set variant); the `DataSet` representation base; updated `Pallet` asset measurements; `BindingState` observation enum.
- Existing types updated where v2.6 / v2.7 changed docstrings or relationship semantics (`AssetChanged` description narrowed; `Configuration.Relationships` now allows asset-to-asset).

### SysML importer — polymorphism-aware emission

- **Drop hand-coded primitive type overrides for `Motion.Axis` / `Motion.Origin` / `CoordinateSystem.Origin` / `Translation.Rotation` / `Translation.Translation` / `SolidModel.Scale`.** The pre-existing `MTConnectPropertyModel.ParseType` switch was flattening polymorphic SysML properties (`AbstractAxis`, `AbstractOrigin`, etc.) into primitive vec3 types (`UnitVector3D`, `Degree3D`). The override pre-dated v2.5's introduction of `*DataSet` siblings; once v2.5 added `<AxisDataSet>` etc. as `<xs:choice>` alternatives to the simple form, the shortcut blocked end-to-end polymorphism. Importer now emits `IAbstractAxis`, `IAbstractOrigin`, `IAbstractRotation`, `IAbstractScale`, `IAbstractTranslation` on `IMotion` / `ICoordinateSystem` / `ITransformation` / `ISolidModel`.
- **Emit all SysML generalizations.** v2.7 declares `OriginDataSet : DataSet, AbstractOrigin` (two `<generalization>` elements per class — same for `AxisDataSet`, `RotationDataSet`, `ScaleDataSet`, `TranslationDataSet`). The pre-existing importer captured only one generalization. Now: `UmlClass.Generalizations[]` collection; heuristic picks the abstract base as the C# class's primary base (since C# does single-inheritance), the rest land as interface inheritance on the matching `I*` declaration (multi-inheritance for interfaces is fine). After the regen: `OriginDataSet : AbstractOrigin, IOriginDataSet, IDataSet`; `IOriginDataSet : IAbstractOrigin, IDataSet`. Same shape across the 5 `*DataSet` siblings. A same-name self-reference filter prevents the v2.7 XMI's `Devices.Device` self-loop from emitting a degenerate `IDevice : IDevice`.
- Parametrize `build/MTConnect.NET-SysML-Import/Program.cs` with `--xmi` / `--output` / `--json-dump` CLI args so the generator runs on Linux / macOS / CI. `--xmi` and `--output` are mandatory; legacy Windows-paths fallback removed.
- Fix two Linux-hostile bugs in the importer (case-sensitive template-path lookups; backslash path-separator) that previously made the generator silently no-op on non-Windows.
- Universal cross-package parent resolver in `MTConnect.NET-SysML`: when a class in one SysML package extends a class in another (e.g. `Devices.Configurations.AxisDataSet` ⇒ `Observation.Representations.DataSet` in v2.7), the resolver grafts the missing parent into the local namespace under a minimal marker form.
- Tighten the `Result`-suffix selector in `CSharpTemplateRenderer` to require a class type before casting.
- Null-guard the importer's model factories, fail-fast on a null model, harden the CSharp template renderer against null/cycle/path edges, and null-guard the Json renderer's `GetDirectoryName` call.
- Null-guard SysML descriptions and properties; harden XMI parsing against malformed input.
- Pin `XmlResolver=null` on the `XmiDeserializer` `XmlDocument` loads.

### Wire format — end-to-end Configuration polymorphism

- New hand-authored wire-format classes for the v2.7 leaf types in `MTConnect.NET-XML/Devices/`, `MTConnect.NET-JSON/Devices/`, `MTConnect.NET-JSON-cppagent/Devices/`:
  - **5 concrete vec3-string elements**: `XmlAxis` / `XmlOrigin` / `XmlRotation` / `XmlScale` / `XmlTranslation` (and Json + Json-cppagent siblings)
  - **5 Entry-keyed DataSet variants**: `XmlAxisDataSet` / `XmlOriginDataSet` / `XmlRotationDataSet` / `XmlScaleDataSet` / `XmlTranslationDataSet`
- Updated parent containers (`XmlMotion`, `XmlCoordinateSystem`, `XmlTransformation`, `XmlSolidModel` and their Json siblings) to handle the polymorphism via parallel optional fields + runtime-type-narrowing dispatch on read; runtime-type dispatch on write.
- cppagent JSON v2 dialect: simple leaves stay inline on the parent (`"Origin": [1.0, 2.0, 3.0]` array form per existing convention); DataSet variants use flat-object PascalCase keys (`"OriginDataSet": {"X": "1.0", "Y": "2.0", "Z": "3.0"}`).

### XML wire format — version-coverage extension

- Extend `libraries/MTConnect.NET-XML/Namespaces.cs` and `Schemas.cs` with the v2.4 / v2.5 / v2.6 / v2.7 namespace URIs and schema-location strings. Pre-fix the maps stopped at v2.5 (Namespaces) and v2.3 (Schemas), so probe / current / sample / asset requests at versions ≥ 2.4 produced HTTP 500 from the formatter when it tried to write a `null` xmlns attribute.

### Test infrastructure

- Rewrite `.github/workflows/dotnet.yml`: fix paths (`src/<lib>/` → `libraries/` + `tests/`), align SDK to 8.0.x + 9.0.x, target `master` (was `main`), upgrade actions to v4, single matrix build-and-test with coverage upload. Trigger restricted to push-to-master + non-draft PRs against master. Pin GitHub Actions to commit SHAs and exclude `XsdLoadStrict` from the default test filter.
- Add `tests/coverlet.runsettings` (shared) + `.config/dotnet-tools.json` (ReportGenerator pin).
- Commit `tools/{dotnet,test}.{sh,ps1}` harness, document the `--docker` flag/env-var dual API, and rename PowerShell `$Args` to `$DotnetArgs` to avoid the automatic-variable shadowing.
- Scaffold three missing paired test projects: `MTConnect.NET-JSON-cppagent-Tests`, `MTConnect.NET-JSON-Tests`, `MTConnect.NET-AgentModule-MqttRelay-Tests`.
- Scaffold the layered compliance harness project `tests/Compliance/MTConnect-Compliance-Tests/` with `L1_XsdValidation/` populated and the `L2_XmiOclAssertions/`, `L3_*` (envelope-shape conformance, reserved), `L4_CrossImpl/`, `L5_Regressions/` layers documented in the project README as not-yet-scaffolded follow-ups.
- Import all 122 official MTConnect XSDs (v1.0 → v2.7) under `tests/Compliance/.../Schemas/v*/`. Test-only fixtures: bundled into the test assembly via `<EmbeddedResource>` and consumed by `SchemaLoadTests` through the manifest API. Not shipped — the compliance project has `<IsPackable>false</IsPackable>` and no library project references the XSDs.
- Add parametric `SchemaLoadTests` that load each XSD via `XmlSchemaSet`. Tagged `[Category("XsdLoadStrict")]`; surfaces 54 failures across 5 root causes (XSD 1.1 features + missing xlink imports — neither resolvable with the .NET BCL validator). The default test filter excludes the category; opt in with `dotnet test tests/Compliance/MTConnect-Compliance-Tests/MTConnect-Compliance-Tests.csproj --filter "Category=XsdLoadStrict"`.
- Add the shared `RepoRootLocator` helper for repo-root resolution in test projects.

### v2.6 / v2.7 unit + compliance + E2E coverage

**Common-Tests (87 cases under `tests/MTConnect.NET-Common-Tests/V2_6_V2_7/`)**:
- `MTConnectVersionsTests` — Version26/Version27 constants, `Max == Version27`, reflection sweep over all 17 versions, no v1.9 constant.
- `V2_6DataItemTypeTests` / `V2_6ComponentAndEnumTests` — v2.6 additions + AssetChanged description regression pin.
- `V2_7DataItemTypeTests` — 8 parametric cases pinning TypeId + Category for every v2.7 DataItem.
- `V2_7ConfigurationDataSetTests` — DataSet base + IDataSet, all `*DataSet` inheritance, concrete leaves, `Abstract*` abstract-modifier preservation.
- `V2_7ComponentTests` — PinTool + ToolHolder.
- `V2_7SampleObservationTests` — round-trip coverage for WaterHardness.

**XML / JSON / JSON-cppagent round-trip (180 cases)** — every Configuration polymorphic leaf round-trips through serialise → deserialise → structural-equality with simple form and DataSet form, plus negative tests for partial-axis machines and `<xs:choice>` violations.

**Compliance — Configuration polymorphism XPath shape (59 cases)** in `tests/Compliance/.../L1_XsdValidation/ConfigurationPolymorphicXsdValidationTests.cs`: 19 polymorphism combinations × 3 versions (v2.5/2.6/2.7) including partial-axis fixtures (Y-only, X+Z-no-Y, A-only, etc.) + 2 negative `<xs:choice>` cases. Validation is structural (XPath), not schema-driven — the v2.5/2.6/2.7 `MTConnectDevices.xsd` files use XSD 1.1 features (notably `<xs:anyAttribute notNamespace="..."/>`) that .NET's XSD-1.0 `XmlSchemaSet` cannot load; pure structural pinning matches the wire-shape requirement without depending on an XSD-1.1 validator.

**E2E — HTTP probe (6 cases)** in `tests/IntegrationTests/Workflows/ConfigurationPolymorphicHttpProbeWorkflowTests.cs`: in-process agent + real `MTConnectHttpClient` issuing `/probe` for each polymorphic combination, asserting structural round-trip end-to-end.

**Workflows catalog**: `docs/testing/workflows.md` rows W08 / W09 / W10 describe the new E2E paths.

Each test cites the authoritative MTConnect Standard source — XMI tag URL/SHA, XSD URL, prose section — at the test fixture or method comment level.

### Documentation

- `docs/testing/v2-6.md` + `docs/testing/v2-7.md` per-version compliance matrices with the new types' status + pinned tests.
- `docs/testing.md` (top-level testing topic doc) + `docs/testing/workflows.md` (workflow catalog) link out to the per-version pages and the harness scripts.
- `build/MTConnect.NET-SysML-Import/README.md` documenting the importer's CLI, the determinism guarantee (zero-diff regen against a pinned XMI tag), the cross-package parent resolver, and an "Adding a new MTConnect Standard version" runbook for v2.8 onwards.
- Root `README.md` + 21 csproj package descriptions advertise "Supports MTConnect versions up to 2.7." (matches the historical wording style; only the maximum version number changes).

### Test result on this PR's tip

Per-project sweep on `feat/issue-133` HEAD with `Category!=RequiresDocker&Category!=XsdLoadStrict`:

| Project | Tests | Pass | Fail |
|---|---|---|---|
| Common-Tests | 87 | 87 | 0 |
| XML-Tests | 73 | 73 | 0 |
| JSON-Tests | 57 | 57 | 0 |
| JSON-cppagent-Tests | 50 | 50 | 0 |
| SHDR-Tests | 27 | 27 | 0 |
| IntegrationTests | 6 | 6 | 0 |
| Compliance | 59 | 59 | 0 |
| **Total** | **359** | **359** | **0** |

### Companion PRs

The wider compliance / coverage / dependency program is split across the following PRs so each one stays independently reviewable; this PR depends on none of them.

- **`chore/deps-update-2026-04-27` (PR #149)** — bumps Scriban 5.9.0 → 7.1.0 to clear 1 critical + 7 high + 3 moderate advisories on the 5.x line, plus the test-infrastructure SDK / NUnit / coverlet bumps Dependabot left out.
- **`test/coverage-and-compliance` (PR #150)** — ships the per-DataItem `MinimumVersion` regen for v2.4-v2.7, fixes the `GenerateDevicesXml` `fileName`-argument bug, makes integration-tests port allocation TIME_WAIT-resilient via OS-assigned ephemeral ports, and lays the scaffold for 100 % unit coverage on regenerated `.g.cs` files, the cppagent parity E2E matrix, round-trip output validation, and full L1-L5 compliance gating.
- **XSD 1.1 + xlink validation** — blocked on validator-strategy input (discussion [#152](https://github.com/TrakHound/MTConnect.NET/discussions/152)). The 54 currently-skipped `SchemaLoadTests` failures cannot be cleared with the .NET BCL `XmlSchemaSet`, which is XSD 1.0 only and does not pre-load the xlink schema; the affected XSDs use XSD 1.1 features (assertions, conditional type assignment) and reference `xlink:href` without an `<xs:import>`. A working solution needs a third-party XSD 1.1 validator (Saxon-HE is the obvious .NET candidate, but other approaches are viable). No PR opened yet — the choice of validator and how to wire it into the L1 harness is being deferred until upstream weighs in on the preferred approach.


ottobolyos:
> I may need to run the SysML import found in the MTConnect.NET-SysML-Import project. But if you are using Claude, etc. I would be curious to see how it would update it on its own.

@PatrickRitchie, I pushed feat/issue-133 to my fork — the v2.6 + v2.7 support is in there end-to-end. Since you asked specifically about how the SysML import got updated to support the new versions, here are the four changes that happened in `build/MTConnect.NET-SysML-Import/`. Posting them here so you can audit before reviewing the PR proper.

### 1. CLI flags (replace hardcoded Windows paths)

The importer used to bake three Windows paths in: `D:\TrakHound\MTConnect\Standard\v2.5\MTConnectSysMLModel.xml`, `C:\temp\mtconnect-model.json`, and an `AppDomain.BaseDirectory`-relative output. CI / Linux / macOS couldn't run it. Replaced with three CLI flags:

```bash
dotnet run --project build/MTConnect.NET-SysML-Import \
    -- --xmi <path-to-MTConnectSysMLModel.xml> \
       --output <repo-root> \
       [--json-dump <path>]
```

`--xmi` and `--output` are mandatory. `--json-dump` is optional (replaces the hardcoded `C:\temp` debug dump). Help text + full guide at [`build/MTConnect.NET-SysML-Import/README.md`](https://github.com/ottobolyos/mtconnect.net/blob/feat/issue-133/build/MTConnect.NET-SysML-Import/README.md) including a "Adding a new MTConnect Standard version" runbook for v2.8 and onwards.

### 2. The F5 flow in Visual Studio — solved with `Properties/launchSettings.json`

After the CLI change, `dotnet run` with no args exits with `error: --xmi <path> is required.` (exit code 2). To keep your "F5 just works" workflow, this PR ships `build/MTConnect.NET-SysML-Import/Properties/launchSettings.json` with three launch profiles. Full file content (committed verbatim):

```json
{
  "$schema": "https://json.schemastore.org/launchsettings.json",
  "profiles": {
    "Import (env vars)": {
      "commandName": "Project",
      "commandLineArgs": "--xmi \"%MTCONNECT_XMI_PATH%\" --output \"%MTCONNECT_NET_REPO%\"",
      "environmentVariables": {
        "MTCONNECT_XMI_PATH": "set this on your machine — absolute path to the SysML XMI file (e.g. C:\\TrakHound\\mtconnect_sysml_model\\MTConnectSysMLModel.xml)",
        "MTCONNECT_NET_REPO": "set this on your machine — absolute path to the mtconnect.net repo root"
      }
    },
    "Import (sibling clone of mtconnect_sysml_model)": {
      "commandName": "Project",
      "commandLineArgs": "--xmi ../../../../mtconnect_sysml_model/MTConnectSysMLModel.xml --output ../../..",
      "comment": "Assumes mtconnect/mtconnect_sysml_model is cloned as a sibling of this repo. Switch standard version with `git -C ../mtconnect_sysml_model checkout v2.7` (or v2.6 / v2.5)."
    },
    "Import (json-dump enabled, sibling clone)": {
      "commandName": "Project",
      "commandLineArgs": "--xmi ../../../../mtconnect_sysml_model/MTConnectSysMLModel.xml --output ../../.. --json-dump ../../../.cache/mtconnect-model.json"
    }
  }
}
```

#### How to use each profile

**A. `Import (env vars)`** — works wherever you keep the SysML model on disk; one-time setup, no edits to checked-in files.

Set the two env vars once on your machine.

```powershell
# Windows (PowerShell), persisted for the user
[Environment]::SetEnvironmentVariable('MTCONNECT_XMI_PATH', 'C:\TrakHound\mtconnect_sysml_model\MTConnectSysMLModel.xml', 'User')
[Environment]::SetEnvironmentVariable('MTCONNECT_NET_REPO', 'C:\source\mtconnect.net', 'User')
```

```bash
# Linux / macOS, persisted in your shell rc file
echo 'export MTCONNECT_XMI_PATH="$HOME/git/mtconnect_sysml_model/MTConnectSysMLModel.xml"' >> ~/.bashrc
echo 'export MTCONNECT_NET_REPO="$HOME/git/mtconnect.net"' >> ~/.bashrc
source ~/.bashrc
```

Then in Visual Studio: drop-down next to the green Run arrow → `Import (env vars)` → F5. Rider: Run / Debug Configuration dropdown → same profile name.

**B. `Import (sibling clone of mtconnect_sysml_model)`** — works without env vars; only requires that `mtconnect/mtconnect_sysml_model` is cloned next to `mtconnect.net`.

```bash
cd $(dirname $(pwd))   # one level up from mtconnect.net
git clone https://github.com/mtconnect/mtconnect_sysml_model
git -C mtconnect_sysml_model checkout v2.7   # or any other tag
# Now the directory layout is:
#   <parent>/mtconnect.net/
#   <parent>/mtconnect_sysml_model/
```

In VS / Rider: select `Import (sibling clone of mtconnect_sysml_model)` → F5. To regen against a different version, `git -C ../mtconnect_sysml_model checkout v2.5` (or v2.6) and press F5 again.

**C. `Import (json-dump enabled, sibling clone)`** — same as B plus a debug-only JSON dump of the parsed model at `<repo-root>/.cache/mtconnect-model.json`. Useful when you suspect the parser is misreading something. The dump path is gitignored via `.cache/` exclusion.

Pick the profile that suits your setup. If you'd prefer a different convention (e.g. absolute paths committed in your fork specifically, or a different env-var contract), trivial to adjust — say the word.

### 3. CI workflow rewrite + `tools/` harness

Two non-importer pieces shipped in this PR that the importer work and every downstream PR rely on. Worth flagging because they touch `.github/workflows/` and the repo-root tooling layout, neither of which I'd otherwise touch in a per-issue PR.

#### `tools/{dotnet,test}.{sh,ps1}` harness

Five small scripts at the repo root (`tools/dotnet.sh`, `tools/dotnet.ps1`, `tools/test.sh`, `tools/test.ps1`, plus the integration-branch refresh helper). The shell ones target Linux/macOS; the `.ps1` siblings target Windows. They wrap the common dev-loop commands (`dotnet build/test/format` with the right flags + the coverlet runsettings + ReportGenerator hook-up). Pure convenience — running `dotnet` directly still works fine.

#### CI workflow rewrite — what was wrong + what changed

The original `.github/workflows/dotnet.yml` had three independent fatal bugs that meant it was effectively a dead workflow on the current tree:

1. **`pull_request: branches: [main]`** — the repo's default branch is `master`, so the filter never matched and PR runs never fired.
2. **`working-directory: src/MTConnect.NET-Common`** (and similar for the other 5 jobs) — actual paths are `libraries/MTConnect.NET-*`. `cd` to a non-existent directory broke every step.
3. **`dotnet test` against a library csproj** — `MTConnect.NET-Common.csproj` is a library, not a test project. `dotnet test` on a non-test project discovers zero tests, so even if (1) and (2) had been right, the workflow ran no actual tests.

The rewritten workflow consolidates the 6 per-library jobs into one solution-level job:

- Triggers: `push: branches: [master]` + `pull_request: branches: [master], types: [opened, synchronize, reopened, ready_for_review]`. Job-level `if: github.event_name == 'push' || github.event.pull_request.draft == false` skips drafts; flipping a draft to ready fires CI immediately on the `ready_for_review` event.
- Action versions: `actions/checkout@v4`, `actions/setup-dotnet@v4` (was `@v2`/`@v1`; v1/v2 are deprecated, Node 20 EOL on Sept 2026).
- SDK: 8.0 + 9.0 (was 7.0; libs target net8/9 — net7 SDK couldn't build).
- `dotnet build MTConnect.NET.sln --configuration Debug` then `dotnet test MTConnect.NET.sln --filter "Category!=RequiresDocker"` — solution-level so it covers every library plus every paired test project.
- Coverage: coverlet via `tests/coverlet.runsettings` → ReportGenerator → upload TRX + cobertura.xml + HTML report as artifact + surface text summary in the GitHub Actions step summary.

Two trade-offs in the rewrite worth flagging because you may want to revisit them:

- **Per-library fail-isolation lost.** The original CI showed `MTConnect-NET-Common: failed` while HTTP/SHDR/etc. continued; now everything is one job. The TRX artifact preserves per-test detail, but if you'd rather have per-library granularity in the GitHub UI, splitting the matrix back out is straightforward.
- **`--configuration Debug` instead of `Release`.** Debug builds only `net8.0`. Release multi-targets all 11 TFMs (`net461..net48`, `netstandard2.0`, `net6.0..net9.0`) — that's what NuGet ships, but the runner doesn't have .NET Framework 4.x dev packs (Linux can't host them) or net6/net7 SDKs. Using Release would fail at restore time on every run. Pragmatic Debug-only fix gives green CI but means a code change that breaks `netstandard2.0` compilation isn't caught until next NuGet publish. Cleaner long-term shape: Debug for PR CI (fast feedback), a separate Release job on push-to-master + scheduled (slow but complete, can use Windows runner for the .NET Framework targets). Happy to layer that in once the foundation is merged.

### Question — keep #139 as one PR, or split it?

The PR is large. v2.6 + v2.7 type-system regen + the test infrastructure + the importer cross-platform fixes + the CI workflow rewrite + the compliance-harness scaffold + the v2.6 / v2.7 unit tests are all interlinked, but each is reviewable in isolation. Three options:

**A. Keep as a single PR.** Pro: one merge; the 9 per-issue PRs (#140–#148) all rebase against this single base, so the rebase happens once. Con: the diff is large (~900 generated `.g.cs` files plus 122 vendored XSDs plus the importer + tests + CI work).

**B. Split into ~5–7 sequential PRs.** A reasonable cut:
1. **`chore/test-infrastructure`** — `.github/workflows/dotnet.yml` rewrite + `tests/coverlet.runsettings` + `.config/dotnet-tools.json` + `tools/{dotnet,test}.{sh,ps1}` harness + paired test-project scaffolding. Lands first; every downstream PR depends on it.
2. **`feat/sysml-import-cross-platform`** — importer `Program.cs` CLI flags + Linux-hostility bug fixes + `Properties/launchSettings.json` + cross-package parent resolver + the wildcard-glob template-copy fix. Generator changes only; no `.g.cs` regen yet.
3. **`feat/v2-6-support`** — `MTConnectVersions.Version26`, regen against `mtconnect/mtconnect_sysml_model@v2.6`, `docs/testing/v2-6.md` matrix, the `V2_6*Tests` fixtures.
4. **`feat/v2-7-support`** — `MTConnectVersions.Version27`, regen against `@v2.7`, `docs/testing/v2-7.md`, the `V2_7*Tests` fixtures (including the `DataSet` representation base + Pallet asset model).
5. **`chore/compliance-harness-scaffold`** — `tests/Compliance/MTConnect-Compliance-Tests/` skeleton with the L1 `SchemaLoadTests` + 122 XSDs imported.
6. **`docs/v2-7-readme-bumps`** — root README + 21 csproj `<Description>` updates to advertise v1.0–v2.7.

Pros: each ~50–200-line diff, easy to review one at a time. Cons: 6 sequential reviews from your queue, vs one big one; the 9 per-issue PRs would each rebase against #1, then #2, etc. as they land — more rebase churn for me, more "is this branch ready yet?" coordination on your end.

**C. Split partial — extract just (1) the CI / tools harness as a tiny prelude PR**, keep the rest in #139. Pro: gets the broken master CI fixed even if #139 takes a long review cycle. Con: #139 is still big; only one extra PR to manage.

Reading your stated preference for one PR per task ([#127 confirmation](https://github.com/TrakHound/MTConnect.NET/pull/127)), the split (B) is closer to your style. But the 9 per-issue PRs already in flight all depend on #139's tree, and re-coordinating their bases against 6 splits would add a meaningful rebase burden. My read: **A or C**, with C as the practical compromise if you want quick CI signal but don't want 6 simultaneous reviews. Happy to execute whichever.

### Documentation strategy — separate discussion

Opened the documentation-site / `docs/index.html` question as its own discussion ([#153](https://github.com/TrakHound/MTConnect.NET/discussions/153)). Out of scope for this PR; a small follow-up will land it once the strategy is decided.

### Determinism guarantee

Critical correctness check that I run after every generator change: **regen against `mtconnect/mtconnect_sysml_model` tag `v2.5` produces zero diff against HEAD.** If a Scriban version, template edit, or parser change shifts output by even a whitespace, this dry-run catches it.

The PR carries:
- v2.6: 20 file changes (4 new types + 16 modified — docstring tweaks, MediaType enum's QIF_MBD addition, etc.)
- v2.7: 89 file changes (incl. the `*DataSet` family + Pallet asset model edits + new DataItems like BindingState/Depth/SwingAngle etc.)

### Why the 122 XSD files are committed (and how to drop them)

The PR adds `tests/Compliance/MTConnect-Compliance-Tests/Schemas/v*/` containing 122 official XSDs (`MTConnectStreams_X.Y.xsd`, `MTConnectDevices_X.Y.xsd`, etc. for v1.0 → v2.7). They sit there as static fixtures consumed by `SchemaLoadTests` and the per-envelope validation matrix that the XSD-validation follow-up PR will turn on. Three reasons they are committed rather than fetched at test time:

1. **Reproducibility.** Tests pin against the exact bytes that were tested. If `schemas.mtconnect.org` ships a corrected XSD for a given version, the test result changes silently — bad for a regression suite.
2. **Offline + air-gapped CI.** `actions/checkout` is the only network step the runner needs. No hard dependency on `schemas.mtconnect.org` being up at test time.
3. **Diff visibility on spec updates.** When a future XSD revision lands, refreshing the fixtures shows up as an explicit diff under `Schemas/`, which the maintainer can review side-by-side with any test-side fix.

The size cost is real: 51.5 MB raw across 122 XML files, ~4 MB after git's pack-time deflate compression. Per-file average is ~420 KB raw / ~33 KB packed — the bulk comes from the count, not from any one outlier.

If you'd rather not vendor them, three alternatives, in order of how invasive each is:

- **A. Fetch-and-cache test setup.** A `[OneTimeSetUp]` hook that downloads each XSD from `schemas.mtconnect.org` into a local cache directory keyed by URL hash. Cache is stored under `~/.cache/mtconnect-xmi-xsd/` (already used for the SysML XMI tags) and committed to `.gitignore`. Tests pin the expected SHA-256 of each fetched file so a server-side change fails the test loudly. CI flakes on `schemas.mtconnect.org` outage.
- **B. Submodule against `mtconnect/schema`.** Tests reference XSDs by relative path inside the submodule. Pin the submodule SHA in the repo. Trades commit-size for submodule maintenance overhead.
- **C. Git LFS for the XSDs.** Keeps the diff-visibility property of the current setup but moves the bytes out of the main pack. Requires `git lfs` on every clone, which is friction for new contributors.

My read: the current vendored copy is the right default for a regression-suite use case. If you prefer (A) for hygiene, easy to wire up in a follow-up PR — say the word and I'll do it.

### Companion work shipping outside this PR

Two items surfaced during this work that don't fit the per-issue scope of #139:

- **`MinimumVersion` annotation gap** — every regenerated DataItem subclass that doesn't carry an explicit version stereotype in the XMI silently falls back to `Version10`. Affects 9 existing DataItems too, not just v2.6 / v2.7 additions. The fix reads the UML stereotype and emits a Scriban-conditional `MinimumVersion` override line. Implemented; ships in a separate companion PR alongside the test-suite expansion.
- **XSD validation across all advertised versions** — 122 official MTConnect XSDs imported; 54 / 122 fail to load via the .NET BCL `XmlSchemaSet` because they use XSD 1.1 features + a missing xlink import. The validator-strategy decision is open as a discussion (#152). Implementation will land as a separate PR once a strategy is agreed. The xlink-import gap is also being filed as a separate report against `mtconnect/schema`.

ottobolyos:
@PatrickRitchie, flipping this one to ready for review. Below is a sequencing checklist for the 13 PRs covering issues #127–#138 + #154 plus a pointer to an integration branch I've been using as a smoke-check.

I've got 13 PRs open against `MTConnect.NET` covering issues #127–#138 + #154. They're independent in scope but their integration order matters for rebase churn. Here's a checklist for sequencing review—tick each off as it lands so we both have visibility on progress.

I've also assembled an integration branch on my fork that has all 13 PRs merged together, so the union builds and tests cleanly end-to-end before any of them lands upstream:

- **Integration branch (all 13 PRs together)**: https://github.com/ottobolyos/mtconnect.net/tree/integration/all-fixes
  - rebuilt by a small refresh script after every per-branch push
  - default `dotnet test` filter is green; `RequiresDocker` E2E suites (Mosquitto + cppagent parity, via Testcontainers) also pass when `MTCONNECT_E2E_DOCKER=true` is exported
  - useful as a smoke-check that the per-PR diffs do not collide once they all land

If you'd like to test-drive the union before approving each PR individually, that branch is the place.

- [x] **PR #139**—`feat/issue-133`—closes **#133**—v2.6 + v2.7 type-system regen + the test infrastructure (`tests/coverlet.runsettings`, `tools/{dotnet,test}.{sh,ps1}` harness, `.github/workflows/dotnet.yml` rewrite, the paired test-project scaffolds). **Land first**—every other PR rebases against its tree.
- [x] **PR #141**—`fix/issue-127`—closes **#127**—`Header.version` emits configured MTConnect release, not library assembly version.
- [x] **PR #145**—`fix/issue-128`—closes **#128**—`schemaVersion` emitted from agent config, not hardcoded literal.
- [x] **PR #143**—`fix/issue-129`—closes **#129**—numeric Sample values emit as JSON numbers (not strings).
- [x] **PR #147**—`fix/issue-130-131`—closes **#130 + #131**—cppagent v2 header fields (`schemaVersion`, `validation`) on every envelope.
- [x] **PR #146**—`fix/issue-132`—closes **#132**—`assetCount` emits in `DATA_SET` representation.
- [x] **PR #159**—`fix/docker-obj-ownership`—out-of-queue tooling fix—the `tools/dotnet.sh --docker` (and transitively `tools/test.sh --docker`) runner now chowns the bind-mounted `bin/` / `obj/` / `TestResults/` / `CoverageReport/` directories back to the host user's UID:GID on container exit. Without this, a docker-mode test run leaves root-owned artefacts on the host that block the contributor's subsequent native `dotnet build`, `rm -rf`, or `git clean`. Independent of every other PR; lands early so the E2E pass on the rest of the queue isn't gated on a sudo round-trip.
- [x] **PR #144**—`fix/issue-134`—closes **#134**—`Organizers.Systems` list refreshed against the System substitution group.
- [x] **PR #140**—`fix/issue-138`—closes **#138**—empty `name` attribute omitted from Probe `DataItem`, `Component`, and `Composition` (`JsonDataItem` / `JsonComponent` / `JsonComposition` skip the assignment when the source `Name` is null or empty). **Reordered to land first of the remaining**—this PR also carries the server-side `IPAddress.TryParse` fix for the Windows `IntegrationTests` flake.
- [x] **PR #150**—`test/coverage-and-compliance`—layered compliance harness (`L1_XsdValidation` / `L4_CrossImpl` / `L5_Regressions`), parametric reflection-driven coverage on the regenerated types, IntegrationTests fixes (thread-safe per-test port allocation, loopback binding, HTTP server startup-exception bubbling), and the `MinimumVersion` SysML stereotype emission that closes the per-DataItem version-introduction-metadata gap. **Reordered to land second**—its IntegrationTests hardening completes the Windows-flake fix begun in #140, so the two land back-to-back ahead of the rest.
- [x] **PR #142**—`fix/issue-135`—closes **#135**—MqttRelay `availability` topic emission corrected.
- [x] **PR #148**—`fix/issue-136-137`—closes **#136 + #137**—`Device` and `Component` ctors no longer back-fill defaults.
- [x] **PR #149**—`chore/deps-update-2026-04-27`—bumps `Scriban` (build-time) + test-infrastructure SDK / NUnit / coverlet versions. Lands late in the queue so package-version conflicts are minimised. Also auto-closes #163 on merge (dependabot's Scriban follow-up, which already superseded #126).
- [x] **PR #155**—`fix/issue-154`—closes **#154**—JSON-cppagent emits `Condition` as the cppagent v2 array-of-wrappers shape (`[{Normal: {...}}, {Warning: {...}}, ...]`) instead of the legacy v1 object-keyed shape (`{Fault: [...], Warning: [...], Normal: [...], Unavailable: [...]}`). Adds a `JsonConverter<JsonConditions>` that writes the v2 shape and reads either form for back-compat. Independent of every other PR in this list.
- [x] **PR #166**—`chore/test-infra-cleanup`—test-infrastructure cleanup sweep that also fixes a real library-side XSD 1.1 incompatibility surfaced during the full-sweep CI restructure. `MTConnect.NET-XML/XmlValidator.cs` now routes through a new `MTConnect.XsdPreprocessor` that strips XSD 1.1 constructs (`xs:assert`, `xs:override`, `notNamespace` on `xs:any` / `xs:anyAttribute`, `xs:any` inside `xs:all`, `maxOccurs > 1` inside `xs:all`) before handing the source to the BCL XSD 1.0 reader, so consumers calling `XmlValidator.Validate` with the official MTConnect XSDs no longer get a list of `Error Adding XSD Schema: notNamespace not supported` diagnostics back. The same helper backs the L1 XSD compliance tests, which are split into 1.0-compatible load (`XmlSchemaSet`) + 1.1-feature presence (`XDocument` parse-and-assert) so #150's broader sweep keeps its coverage. Also includes a `RequiresDocker` / `E2E` path-agnostic refactor (no host-path coupling under Docker-in-Docker), a `coverlet.collector 3.2.0 → 6.0.4` per-csproj bump (the 3.2.0 line hangs in `Mono.Cecil.Rocks` on generated `.g.cs` DLLs), and a `dotnet.yml` matrix-include restructure that runs `RequiresDocker` + `E2E` + `XsdLoadStrict` with no filter on the ubuntu leg and keeps `Category!=RequiresDocker&Category!=E2E` on the windows leg (with explanatory comment) since hosted Windows runners can't host Linux Docker. **Lands first among the remaining**—it's the test-infrastructure prerequisite #165's full-sweep CI depends on, and the library-side XSD 1.1 fix is consumer-visible enough to deserve its own ahead-of-everything slot.
- [x] **PR #165**—`fix/json-cppagent-mqtt-devices-envelope`—fixes a high-severity defect in the `JSON-cppagent-mqtt` Probe formatter: `JsonMqttResponseDocumentFormatter.Format(IDevicesResponseDocument, …)` serialised only the first device wrapped in a `JsonDeviceContainer` instead of the full `MTConnectDevices` envelope, so every MQTT Probe topic dropped all but the first device and every machine `<Device>` was silently lost when the agent meta-device sat at the head of the list. The fix routes the MQTT path through `JsonDevicesResponseDocument` exactly like the sibling HTTP formatter — `JsonDevices` already dispatches on `device.Type == Agent.TypeId` to separate Agent[] and Device[] keys inside `Devices`, so no new dispatch logic is needed. **Lands second**—after #166 lands, #165's full-sweep CI runs cleanly under the matrix-include filter without the XSD 1.1 / DinD / coverlet blockers.
- [x] **PR #167**—`fix/json-mqtt-probe-envelope-test-race`—test-only hotfix for a concurrent-mutation race in the `Probe_envelope_multi_device_round_trip_preserves_every_device` integration test that came in via #165's merge. The assertion path read the `received` `Dictionary<string, byte[]>` outside `lock (receivedLock)` while the MQTT subscriber handler could still mutate it from a late QoS-1 retransmit, which `ubuntu-latest` CI trips reliably under load (the local Docker-in-Docker loop does not). Fix snapshots `received.ToArray()` once under the lock and routes the assertion message + foreach through the snapshot; the handler itself is unchanged. **Merged 2026-05-26** (commit `6911622d`).
- [x] **PR #168**—`fix/agent-uuid-config-override`—spec-conformance fix for the per-boot UUID regeneration in `MTConnectAgentApplication.StartAgent`. Adds `AgentUuid` to `IAgentApplicationConfiguration` + `AgentApplicationConfiguration` (with `[JsonPropertyName("agentUuid")]`); threads it through `StartAgent` to override `agentInformation.Uuid` before the durable/InstanceId-reset block. When neither operator config nor a persisted state file supplies a UUID—the typical ephemeral-container first-boot case—falls back to a deterministic UUID v5 derived per RFC 4122 §4.3 from the DNS namespace and `agentName:port`, mirroring cppagent's `name_generator` so the agent meta-device satisfies `UuidType`'s 'for it's entire life' annotation across every restart out of the box. 13 NUnit tests pin the contract end-to-end (3 override-mechanic + 3 longitudinal + 3 deterministic-default + 1 RFC-vector check + 3 misc). **Merged 2026-05-27** (commit `f199be22`)—master now at `f199be22`.
- [x] **PR #169**—`fix/device-ctor-no-random-defaults`—startup-validation hook that completes the device-identity contract that `#136`/`#137` (PR #148) began. The parameterless `Device()` ctor was already cleaned up at master (commit `38fa328e`) so `Id` / `Name` / `Uuid` are all `null` on construction; this PR adds an `MTConnectAgent.AddDevice` gate that throws `InvalidOperationException` with the device's registration index + type + name (and the XSD citation) when `Uuid` is null or empty. Closes the path where a missing config knob would surface as an opaque `ConcurrentDictionary.ArgumentNullException` far from the cause. **Merged 2026-05-27** (commit `a57d5fa5`)—master now at `a57d5fa5`.
- [x] **PR #170**—`fix/agent-instanceid-persist-nonzero`—spec-conformance fix in `MTConnectAgentApplication.StartAgent` for the perpetual `instanceId = 0` reset. Replaces the bare `agentInformation.InstanceId = 0` assignment (which broke `InstanceIdType`'s `xs:minInclusive value='1'` for every file-reading consumer and defeated the persist-and-restore contract on same-process restart cycles) with `agentInformation.InstanceId = Math.Max(agentInformation.InstanceId + 1, (ulong)UnixDateTime.Now)`. The `+1` defeats same-second restart collisions (the SysML XMI MUST-clause: *'MUST be changed to a different unique number each time the buffer is cleared'*) while the `Max` floor keeps the value time-meaningful. Closes three defect surfaces at once—XSD schema validity, persist-and-restore symmetry, strict per-restart uniqueness—across both the persisted file and the live envelope path. 4 NUnit tests. **Merged 2026-05-27T20:31Z** (commit `f3200b2b`)—master now at `f3200b2b`.
- [x] **PR #162**—`chore/sysml-model-submodule`—build-infrastructure chore—adds `mtconnect_sysml_model` as a tracked git submodule at `build/sysml-model` (pinned to tag `v2.7`), replacing the documented-but-untracked external-clone pattern the `*.g.cs` regen pipeline relied on. It's independent of every code PR and nothing in the queue depends on the submodule path yet, so it can land at any point; it's grouped with the chores ahead of the final clean-up. **Merged 2026-05-28T00:45Z** (commit `4c49765b`)—master now at `4c49765b`.
- [x] **PR #160**—`chore/scrub-section-sigil`—campaign clean-up sweep—replaces the literal `§` character with the word `section ` across ten committed files that landed via the already-merged half of the queue. The character is harmless in spec citations (`Part 1.0 §3 Header`) but doubles as the tripwire for the local repo-side rule-book grep, generating false positives on every audit run. Lands last among the in-place clean-ups so the rebased history surface settles before the cleanup applies. **Merged 2026-05-28T02:18Z** (commit `8ac2185f`)—master now at `8ac2185f`.
- [x] **PR #164**—`chore/warning-cleanup-2026-05-22`—drives every `dotnet build --no-incremental` warning to zero across every code family (CS / CA / SYSLIB / NU), then locks the gain in place with `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` in `Directory.Build.props` so a future regression breaks the build. Baseline 2026-05-22 was 1,114 warnings (top families: CS0436 type-collision 452, CS8632 nullable-context 220, CS8603 null-return 78, CS0618 obsolete-API 68). One commit per code family for review-by-chunk. Lands after #160 and before #157 so the regression gate sits on the campaign tail and #157's docs-build inherits the zero-warning baseline. **Merged 2026-05-28T15:10Z** (commit `6f2e1a3b`)—master now at `6f2e1a3b`.
- [x] **PR #172**—`chore/remove-http-aspnetcore`—removes the obsolete `MTConnect.NET-HTTP-AspNetCore` library. The `AgentModule-HttpServer` replacement has been the recommended HTTP transport since it landed; `MTConnect.NET-HTTP-AspNetCore` is unmaintained, its API surface overlaps the newer module's, and discussion #161 recorded Patrick's explicit approval to drop it. Lands at the chore tail, after #164's zero-warning regression gate and before #157 so the docs site never generates documentation for a library that will no longer exist. **Merged 2026-05-28T16:39Z** (commit `478621a5`)—master now at `478621a5`.
- [x] **PR #157** `feat/vitepress-docs` — DRAFT (rebuilt 2026-05-27 atop integration tip as 29-commit clean atomic history; train tail per "docs pr must be the last")

Code surfaces are mostly disjoint between the per-issue PRs. Rebasing each against `master` after a sibling merges is a no-op in the typical case.

**Reordered 2026-05-18**: I've moved #140 and #150 to the front of the remaining queue. Together they close the Windows `IntegrationTests` flake—#140 carries the server-side `IPAddress.TryParse` fix and #150 carries the test-infrastructure hardening (thread-safe per-test port allocation, loopback binding, HTTP-server startup-exception bubbling). While they sit behind #142/#148, every PR ahead of them shows a red `build-and-test-windows-latest` leg purely from the catalogued `ClientAgentCommunicationTests` timeout, not from anything those PRs touch. Landing #140 then #150 first turns that leg green for the rest of the queue. The remaining order after the two is #142 → #148 → #149 → #155 → #160; the merged half stays as it landed.

**Reordered 2026-05-22**: With #150 + #140 + #142 landed and the Windows-flake fixes complete, the remaining order is #148 → #149 → #155 → #162 → #160 → #164 → #157. The chore-tail (#162 sysml-submodule, #160 sigil-scrub, #164 warning-cleanup) sits ahead of #157 so the docs site sees the final post-campaign surface with the zero-warning regression gate already in place.

**Reordered 2026-05-23**: With #148 + #149 + #155 landed and a new functional bug surfaced in the `JSON-cppagent-mqtt` Probe formatter, the remaining order is #165 → #162 → #160 → #164 → #157. #165 sits at the head because it's a structural defect in the JSON v2 Probe envelope (multi-device payloads collapsed to the first device, agent meta-device erasing every `<Device>`); the chore-tail (#162 sysml-submodule, #160 sigil-scrub, #164 warning-cleanup) sits behind it ahead of #157 so the docs site continues to see the final post-campaign surface.

**Reordered 2026-05-24**: A new chore PR—#166 `chore/test-infra-cleanup`—jumped to the head of the remaining queue. It carries a real library-side BCL XSD 1.1 fix (`XmlValidator.Validate` now preprocesses via the new `MTConnect.XsdPreprocessor` before `XmlSchema.Read`) plus the test-infrastructure prerequisites the full-sweep CI on #165 depends on—the XsdLoadStrict pattern restore using the same shared preprocessor, the `RequiresDocker` / `E2E` path-agnostic refactor, the `coverlet.collector 6.x` bump, and the `dotnet.yml` matrix-include. The remaining order is now #166 → #165 → #162 → #160 → #164 → #157.

## Open questions blocking related follow-up work

Three items I'd like your steer on before opening the corresponding PRs—each is its own discussion thread so we can converge separately:

- **XSD 1.1 / xlink-aware envelope validation**—discussion [#152](https://github.com/TrakHound/MTConnect.NET/discussions/152). 54 of 122 published MTConnect XSDs need an XSD 1.1 validator; .NET BCL only does XSD 1.0. I'm asking which library / approach you'd accept.
- **Documentation site / `docs/index.html`**—discussion [#153](https://github.com/TrakHound/MTConnect.NET/discussions/153). The 12-byte `HELLO WORLD` placeholder suggests an unfinished docs-site setup. I'm asking whether you'd like a static site (VitePress / Docusaurus / DocFX), a `docs/index.md` hub page, or just deletion.
- **Generator improvements + code-quality cleanup + performance pass**—three internal-hygiene plans that span the codebase (e.g. wire `MaximumVersion` deprecation marking into the SysML importer; resolve the `IPNetwork` BCL collision + the obsolete MQTTnet TLS API call sites; cache hot-path lookups). Each would land as its own PR. I haven't opened them because they're 'would you like this?' rather than 'fix this issue'—I'd rather scope each with you first than push speculative work. I'd be happy to draft separate discussions for each if useful, or to tackle them piecemeal as you greenlight individual items.

## What this isn't

- It isn't a strict requirement—any merge order produces the same end state.
- It isn't a deadline ask—I'm happy to follow your review cadence.
- It isn't coupled to any review tooling on my end—every PR is independently reviewable on GitHub's web view.

If a different order or batching shape works better for your queue, just say.

**Reordered 2026-05-24 (afternoon)**: With #166 merged (commit `2872e4c7`), the chore-tail clean-up sweep + library-side XSD 1.1 fix + test-infra prerequisites are in. The remaining queue is #165 → #162 → #160 → #164 → #157. #165 jumps to the head — its CI now runs the full sweep (all categories) thanks to #166's matrix-include, and the runtime-type / observation-factory fixes it carries are independently bluefin-verified.

**Reordered 2026-05-24 (evening)**: With #165 merged (commit `1be96a1e`), the JSON-cppagent envelope bug-fix is in. The remaining queue is #162 → #160 → #164 → #157. #162 jumps to the head — Patrick has not yet reviewed it, so auto-flip is on hold pending review or explicit user override.

**Reordered 2026-05-24 (late evening)**: A new test-only hotfix PR—#167 `fix/json-mqtt-probe-envelope-test-race`—jumps to the head of the remaining queue. It carries a one-region fix for a concurrent-mutation race in `JsonMqttProbeEnvelopeIntegrationTests` that came in via #165's merge: the assertion path read a `Dictionary<string, byte[]>` outside `lock (receivedLock)` while the still-live MQTT subscriber handler could mutate it on a late QoS-1 retransmit, which the `ubuntu-latest` CI runner trips reliably under load (the local Docker-in-Docker loop does not). The remaining queue is now #167 → #162 → #160 → #164 → #157; the chore tail keeps its existing relative order, and #162's currently-red CI should turn green on the no-op cascade-rebase once #167 lands.

**Reordered 2026-05-25**: Three new fix PRs—#168 (agent-uuid config override), #169 (device-ctor startup validation), #170 (instanceid non-zero reset)—jump into the queue immediately after #167. All three are spec-conformance fixes against the MTConnect v2.7 XSD: `UuidType` lifetime identity, `Device.uuid use='required'`, and `InstanceIdType xs:minInclusive value='1'` respectively. They land before the chore tail (#162 / #160 / #164) and the docs PR (#157). The remaining queue is now #167 → #168 → #169 → #170 → #162 → #160 → #164 → #157.

**Merged 2026-05-27**: With #168 merged (commit `f199be22`, master tip `f199be22`), the agent-UUID config-override spec-conformance fix is in. The remaining queue is #169 → #170 → #162 → #160 → #164 → #157. #169 jumps to the head—it's now ready for review; all six remaining branches have been cascade-rebased onto master `f199be22` and force-pushed. Latest patch: 2026-05-27T00:30Z.

**Reordered 2026-05-27**: Inserted #172 (`chore/remove-http-aspnetcore`) into the chore tail immediately before #157, per the maintainer's discussion-#161 removal approval. The remaining queue is now #169 → #170 → #162 → #160 → #164 → #172 → #157.

**Rebuilt 2026-05-27**: #157 (`feat/vitepress-docs`) has been fully rebuilt as a 29-commit clean atomic history atop the current `integration/all-fixes` tip and force-pushed (`189a4b61` -> `53b0ae40`). It rejoins the train at the tail per the standing rule ("docs pr must be the last"). The remaining queue is unchanged: #169 → #170 → #162 → #160 → #164 → #172 → #157.

**Merged 2026-05-27 (evening)**: With #169 merged (commit `a57d5fa5`, master tip `a57d5fa5`), the device-ctor startup-validation hook is in. The remaining queue is #170 → #162 → #160 → #164 → #172 → #157. #170 jumps to the head—all six remaining branches have been cascade-rebased onto master `a57d5fa5` and force-pushed (clean cherry-pick on each, debug builds green).

**Merged 2026-05-27 (late evening)**: With #170 merged (commit `f3200b2b`, master tip `f3200b2b` at 20:31Z), the agent-InstanceId-persist-nonzero spec-conformance fix is in. The remaining queue is #162 → #160 → #164 → #172 → #157. #162 jumps to the head—all five remaining branches have been cascade-rebased onto master `f3200b2b` and force-pushed (clean cherry-pick on each, debug builds green). #160 is APPROVED but waits behind #162 per the at-most-one-ready invariant; it flips to ready immediately after #162 lands.

**Merged 2026-05-28**: With #162 merged (commit `4c49765b`, master tip `4c49765b` at 00:45Z), the SysML-model submodule chore is in. The remaining queue is #160 → #164 → #172 → #157. #160 jumps to the head—all four remaining branches have been cascade-rebased onto master `4c49765b` and force-pushed (clean cherry-pick on each, debug builds green). #160 is already APPROVED and flips to ready immediately per the at-most-one-ready invariant.

**Merged 2026-05-28 (post-#160)**: With #160 merged (commit `8ac2185f`, master tip `8ac2185f` at 02:18Z), the section-sigil scrub clean-up is in. The remaining queue is #164 → #172 → #157. #164 jumps to the head—all three remaining branches have been cascade-rebased onto master `8ac2185f` and force-pushed (clean cherry-pick on each, debug builds green). #164 flips to ready immediately per the at-most-one-ready invariant.

**Merged 2026-05-28 (post-#164)**: With #164 merged (commit `6f2e1a3b`, master tip `6f2e1a3b` at 15:10Z), the warning-cleanup campaign and its `TreatWarningsAsErrors` regression gate are in. The remaining queue is #172 → #157. #172 jumps to the head—both remaining branches have been cascade-rebased onto master `6f2e1a3b` and force-pushed (clean cherry-pick on each, debug builds green at zero warnings under the new gate). #172 flips to ready immediately per the at-most-one-ready invariant.

**Merged 2026-05-28 (post-#172)**: With #172 merged (commit `478621a5`, master tip `478621a5` at 16:39Z), the AspNetCore-removal chore is in. That lands the entire bug-fix + chore train—every code and infrastructure PR in this checklist is now on master. The remaining queue is #157 only. #157 (`feat/vitepress-docs`) is the campaign's final deliverable; it has been rebased directly onto master `478621a5` (the integration branch now equals master, since the in-flight train is empty), force-pushed, and flipped to ready. Once it merges, the campaign closes.

ottobolyos:
@PatrickRitchie — landed a few additional fixes since the original ready-for-review flip a couple of days ago that improve the foundation this PR ships:

- **SysML importer Linux-portability** — the importer was silently no-op'ing on case-sensitive filesystems (lowercase `csharp/templates/` paths against the on-disk `CSharp/Templates/`, literal `\\` separators, hardcoded `C:\temp\` output). Now portable across Linux / macOS / Windows via `Path.DirectorySeparatorChar` + case-correct paths + a CLI-driven `--xmi`/`--output` flag. Without this, every regen on a CI runner produced no output.
- **Multi-generalization importer** — `UmlClass.Generalization` (singular) became `UmlClass.Generalizations[]` (collection) so the importer can capture the SysML's multi-inheritance shape. This is what unblocked the v2.7 polymorphism work (`OriginDataSet : AbstractOrigin, IDataSet`, etc.) — without it the second generalization was silently lost and the wire-format polymorphism couldn't compile.
- **`ResolveDanglingParents`** — when a class's generalization points into a sibling SysML package the local parser does not walk, the resolver grafts a structurally-minimal C# base into the local namespace so the generated children compile. Version-agnostic — fires only when there's a dangling parent, so older XMIs are no-ops.
- **A handful of generator hardcodes flagged + fixed** — Result-Class TABLE-vs-DATA_SET classifier, Interface DataItem REQUEST/RESPONSE subtypes enum, ControllersComponent `MinimumVersion = v2.0` per the SysML's `introduced='2.0'`. Each was a small one-liner in the importer that produced surprising `.g.cs` defaults until corrected.

PR #139 is ready for review. The merge-order checklist comment above stays current — if you start with this PR and follow the rest in any order, the rebase-against-`master` cost between siblings is near zero in practice.

PatrickRitchie (approved):
Looks great!

PR #141, opened 2026-04-25
fix: emit configured mtconnect release in header.version (#127)

### LINKED ISSUE #127: `Header.version` reports library assembly version instead of MTConnect release
## Summary

Every envelope published by the `mqtt-relay` module (and, by extension, any consumer of the shared Header serialiser) reports `Header.version` as the library's own assembly version (e.g. `"6.9.0.0"` on `v6.9.0.2`), rather than the MTConnect Standard release the agent is serving. This misidentifies the agent to every downstream consumer and breaks any logic that branches on spec version.

## Verified by reproduction — 2026-04-20

Reproduced live against `trakhound/mtconnect.net-agent:6.9.0` (the official TrakHound image published 2025-10-16 alongside v6.9.0.2) with `eclipse-mosquitto:2.0.22` as the broker. The `Header.version` slot emits `"6.9.0.0"` on every `MTConnectStreams` MQTT envelope, and `version="6.9.0.0"` on the HTTP XML Probe Header, regardless of any `defaultVersion` value set in `agent.config.yaml`. Side-by-side comparison: cppagent v2.7.0.7 in an identical configuration emits `"2.7.0.7"` in the same slot — correctly reporting the agent's MTConnect release.

## Environment

- `MTConnect.NET-Applications-Agents` + `MTConnect.NET-SHDR` + `MTConnect.NET-JSON-cppagent` at `v6.9.0.2` (published 2025-10-16, the latest release).
- Format ID: `JSON-cppagent-mqtt` (the same shared Header serialiser is also used for `JSON-CPPAGENT` — please confirm the scope when fixing).

## Reproduction

`applications/MTConnect-Agent/appsettings.yaml` (minimum viable):

```yaml
defaultVersion: 2.5.0.0       # the library's advertised ceiling; any valid value surfaces the same bug
modules:
  - mqtt-relay:
      server: localhost
      port: 1883
      topicPrefix: MTConnect
      documentFormat: JSON-cppagent-mqtt
```

Start a local Mosquitto (`eclipse-mosquitto:2.0.22` on `127.0.0.1:1883`), run the agent, and subscribe:

```sh
mosquitto_sub -h localhost -t 'MTConnect/Current/#' -C 1 | jq '.MTConnectStreams.Header.version'
```

### Observed

```json
"6.9.0.0"
```

### Expected

```json
"2.5.0.0"
```

(With `defaultVersion` unset, the companion hardcoded-`schemaVersion` issue #128 further forces this to `"2.0"`; with `defaultVersion` correctly wired to the configured release, it should echo that release.)

## Authority

`MTConnectStreams_2.7.xsd` (identical wording in `_2.5.xsd` and `_2.6.xsd`):

```xml
<xs:attributeGroup name='HeaderAttributesType'>
  <xs:attribute name='version' type='VersionType' use='required'>
    <xs:annotation>
      <xs:documentation>The document version</xs:documentation>
    </xs:annotation>
  </xs:attribute>
  ...
</xs:attributeGroup>

<xs:simpleType name='VersionType'>
  <xs:restriction base='xs:NMTOKEN'/>
</xs:simpleType>
```

The XSD does not pattern-constrain `VersionType`, but the MTConnect prose standard, Part 1 (Overview, §5) and Part 2 (Devices Information Model, §7 Header), documents `version` as the **MTConnect Standard release** the agent serves (e.g. `"2.5.0.0"`, `"2.7.0.0"`).

The [cppagent reference implementation](https://github.com/mtconnect/cppagent) emits exactly that — `"2.7.0.0"` on a v2.7 agent, `"2.5.0.0"` on a v2.5 agent. The format ID `JSON-CPPAGENT-MQTT` is itself an explicit promise of cppagent-reference compatibility; diverging here breaks both the prose-spec contract and the format-label contract.

## Impact

- Consumers that branch on `Header.version` to decide whether v2.x-only fields should be surfaced receive a version string they can never match against any published MTConnect release.
- Strict cppagent-shape consumers fail format-label validation.
- Operational diagnostics (e.g. schema registries, compliance dashboards) misreport the agent as running an undefined "6.9.x" MTConnect version.

## Location in source

Shared Header serialiser used by `libraries/MTConnect.NET-JSON-cppagent/Formatters/JsonMqttResponseDocumentFormatter.cs`. The encoder currently reads the MTConnect.NET assembly version when populating the `version` slot instead of the configured `DefaultVersion` / agent release.

## Suggested fix

1. **Emit the configured MTConnect release** as `Header.version` — the same value used for `schemaVersion`. Library assembly version should be exposed via metrics, a non-standard extension field, or a `sender` suffix, but never in the spec-defined `version` slot.
2. If the value is currently pulled from `Assembly.GetExecutingAssembly().GetName().Version`, switch to reading the agent's configured MTConnect version (`AgentApplicationConfiguration.DefaultVersion` or equivalent).

Identical across v2.5, v2.6, v2.7 — `HeaderAttributesType.version` has been a required attribute typed `VersionType` since v1.3+, and the prose-spec guidance that this reflects the MTConnect release has been stable across every v2.x release.

## Related issues

- #128 — *`MTConnectStreams.schemaVersion` / `MTConnectDevices.schemaVersion` hardcoded to `"2.0"`* — same envelope, adjacent root cause; the two slots should report the same MTConnect release.
- #133 — *Add MTConnect v2.6 + v2.7 support* — once available, the fix here automatically serves the latest release rather than the library's compiled-in ceiling.

## References

- XSDs: [`MTConnectStreams_2.7.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectStreams_2.7.xsd), [`MTConnectStreams_2.6.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectStreams_2.6.xsd), [`MTConnectStreams_2.5.xsd`](https://github.com/mtconnect/schema/blob/master/MTConnectStreams_2.5.xsd).
- Prose standard: <https://www.mtconnect.org/standard-download20181> — Part 1 §5, Part 2 §7.
- Normative SysML / XMI model: <https://github.com/mtconnect/mtconnect_sysml_model> — `MTConnectSysMLModel.xml`.
- cppagent reference: <https://github.com/mtconnect/cppagent>.

ottobolyos:
## Summary

Fixes #127 — `Header.version` emitted the library assembly version (`6.9.0.0`) instead of the configured MTConnect Standard release.

- Route the configured MTConnect release (`MTConnectAgent.MTConnectVersion`) into `Header.Version` for every envelope produced by `MTConnectAgentBroker` (Devices, Streams, Assets, Error). A new private `FormatHeaderVersion` helper emits the four-segment shape (e.g. `2.5.0.0`) so the value round-trips with the cppagent reference.
- Cache the formatted `Header.Version` per release so repeated envelope construction does not re-format the same `MTConnectVersions` constant.
- Drop the six redundant `header.Version = Version.ToString()` overwrites in the response-document construction methods now that the four header builders carry the correct value.
- Add an optional `Version mtconnectVersion` parameter to the previously-parameterless `GetErrorHeader` so the two `GetErrorResponseDocument` overloads route the configured release through the Error envelope as well.
- Use C# 7.3-compatible syntax in the broker so the file compiles on the legacy TFMs (`netstandard2.0` / `net48`).
- Keep `MTConnectAgent.Version` (library assembly version) as a legitimate diagnostic surface; public API unchanged.
- Add NUnit regression coverage pinning the new behavior across every `MTConnectVersions` constant for Devices / Assets / Error envelope kinds plus the explicit-version override, a repo-wide guard that rejects any wire-format envelope containing the library assembly version, and XML round-trip cases asserting the wire shape via `XmlDevicesResponseDocument.ToXmlStream`.


PatrickRitchie (approved):

ottobolyos:
Flipping this one to ready for review now that #139 has landed. Branch rebased onto current master; the rebase replayed cleanly with no conflicts and the prior approval carried over to the new tip.

Per the sequencing checklist on #139, this is item 2 of 13. I'll hold #145 (`fix/issue-128`) in draft until this one merges so each PR goes in one at a time and the queue stays tidy.

PatrickRitchie:
Ok sounds good. I will continue to review the PRs following the sequence in #139

PR #142, opened 2026-04-25
fix(agent-module): correct MqttRelay availability topic emission

### LINKED ISSUE #135: `mqtt-relay` module publishes Agent Availability on a non-canonical `{prefix}/Probe/{agent-uuid}/Available` four-segment topic

ottobolyos:
## Summary

Fixes #135 — the MqttRelay agent module published Agent Availability on `{TopicPrefix}/Probe/{agent-uuid}/Available` with a raw `AVAILABLE` / `UNAVAILABLE` string, breaking the contract that `{TopicPrefix}/Probe/#` carries only JSON envelopes.

- Relocate the Availability last-will-and-testament topic and the on-connect retained-message topic to `{TopicPrefix}/Agent/{agent-uuid}/Available`, outside the Probe wildcard.
- Validate `AvailabilityTopic` configuration inputs and reject empty / whitespace / wildcard-segment values before the relay starts.
- Guard `MqttRelay.OnStop` against a NullReferenceException when the relay shuts down in Entity mode.
- Bound the shutdown disconnect await so a stuck broker cannot hang Agent termination.
- Back the last-sent-sequence persister with an in-memory store + timed flush that holds the counter under `Interlocked.Read` / `Interlocked.Exchange`, so concurrent observation publishing cannot tear the 64-bit value or thrash the I/O path.
- Group observations in a single pass during MqttRelay publish so per-tick CPU stays linear in the observation count.
- Wrap the relay's async-void event handlers with a fault-logging shim and log unexpected Worker faults from the outer catch so silent dead-thread regressions surface in the agent log.
- Guard the missed-observation counter against integer underflow when a sequence reset arrives before the persister has flushed.
- Add `MTConnect.NET-AgentModule-MqttRelay-Tests` with NUnit coverage of the topic-construction helper, a structural guard refusing any future availability topic containing `/Probe/`, and unit pins for the shutdown / fault-handling / atomic-counter / single-pass-grouping / in-memory-persister contracts.
- Document the operator migration path in `agent/Modules/MTConnect.NET-AgentModule-MqttRelay/README-Nuget.md` (and the in-tree `README.md`).
- Breaking: operators that subscribe to `{TopicPrefix}/Probe/{agent-uuid}/Available` for the raw-string availability state must move their subscription to `{TopicPrefix}/Agent/{agent-uuid}/Available`. Subscribers of the `Probe/#` wildcard now receive only JSON envelopes.


PatrickRitchie (approved):

ottobolyos:
@PatrickRitchie, I've put this PR back in draft and reordered the queue—here's the reasoning.

The only red check here is the catalogued Windows `IntegrationTests` flake, not anything this PR touches. `build-and-test-windows-latest` fails on `IntegrationTests.ClientAgentCommunicationTests.WaitForSampleShouldSucceedAfterFirstItemIsSent` with `System.TimeoutException: HTTP Stream Timeout Exceeded (3000)`; `build-and-test-ubuntu-latest` is green, and every other test assembly passes on Windows too. This PR's diff is MqttRelay `availability` topic emission only; the failing test exercises the client↔agent HTTP path and is unrelated.

I've moved #140 and #150 to the front of the remaining queue, ahead of this PR. Together they close that flake—#140 carries the server-side `IPAddress.TryParse` fix and #150 carries the test-infrastructure hardening (thread-safe per-test port allocation, loopback binding, HTTP-server startup-exception bubbling). Once the two land, this PR rebases onto a fixed `master` and the Windows leg goes green naturally, so I've kept #142 in draft until then rather than ask you to merge through a red matrix. The sequencing checklist on #139 reflects the new order.

Nothing in this PR's scope changed: the remaining order is #140 → #150 → #142 → #148 → #149 → #155 → #160.

PatrickRitchie (approved):

ottobolyos:
@PatrickRitchie, is there anything you’d like me to change in this PR? 🤔

PatrickRitchie:
@ottobolyos Looks good! I was just waiting on #150 before I merged this as that was the sequence you mentioned

ottobolyos:
You’re right, my bad. You can review #150 now. :wink:

PatrickRitchie (approved):

PR #156, opened 2026-04-30
fix(common): expose AssetBufferSize and AssetCount on Streams Header

ottobolyos:
## Summary

`MTConnect.NET-JSON-cppagent` Streams envelope `Header` omits the `assetBufferSize` and `assetCount` attributes that the cppagent JSON v2 reference printer emits on every envelope's Header. Per the XSD's `HeaderStreamsAttributesType` (which derives from `HeaderAttributesType`), both attributes are inherited and optional — the cppagent reference always emits them.

`IMTConnectDevicesHeader` and `IMTConnectAssestsHeader` already exposed these fields. Only the Streams Header was missing the surface, surfacing as a wire-format omission on every Streams envelope.

## Reproduction

Boot a `JSON-cppagent-mqtt` agent with `assetBufferSize: 8192` configured. Capture a Current envelope:

```bash
mosquitto_sub -t 'MTConnect/Current/+' -h $BROKER -C 1 | jq '.MTConnectStreams.Header'
```

Pre-fix output:
```json
{
  "instanceId": 1776878246,
  "version": "6.9.0.0",
  "sender": "agent",
  "bufferSize": 131072,
  "firstSequence": 1,
  "lastSequence": 76,
  "nextSequence": 77,
  "deviceModelChangeTime": "...",
  "creationTime": "..."
}
```

cppagent v2.7.0.7 in an identical configuration emits both `assetBufferSize` and `assetCount`.

## Consumer impact

Downstream consumers monitoring multi-agent fleets cannot read asset-buffer utilisation from Current/Sample envelope Headers — only from less-frequent Asset envelopes. Strict cppagent JSON v2 parsers default the missing fields to `0` to keep parsing succeeding, but the resulting accounting rows under-report the agent's actual asset buffer state.

## Fix

- Add `AssetBufferSize` and `AssetCount` to `IMTConnectStreamsHeader` and `MTConnectStreamsHeader` (matches the existing properties on `IMTConnectDevicesHeader` / `IMTConnectAssestsHeader`).
- Add the matching JSON-cppagent serialisation (`assetBufferSize`, `assetCount`) on `JsonStreamsHeader`, including the `ToStreamsHeader()` round-trip path.
- Wire the producer-side population in `MTConnectAgentBroker.GetStreamsHeader`, sourcing values from the existing `_assetBuffer` field (matches the parallel `GetAssetsHeader` path).

## Tests

`tests/MTConnect.NET-Common-Tests/Headers/MTConnectStreamsHeaderAssetFieldsTests.cs` pins:

1. `IMTConnectStreamsHeader` declares `AssetBufferSize` getter (typed `ulong`).
2. `IMTConnectStreamsHeader` declares `AssetCount` getter (typed `ulong`).
3. `MTConnectStreamsHeader` round-trips both fields through get/set.
4. Default values are `0` (no buffer configured / no assets).

4/4 pass.

## Spec authority

- XSD: `https://schemas.mtconnect.org/schemas/MTConnectStreams_2.7.xsd` — `HeaderStreamsAttributesType` derives from `HeaderAttributesType` which carries both attributes.
- Reference: cppagent v2.7.0.7 `printer/json_printer.cpp` emits both fields on every envelope's Header.
- Prose: MTConnect Standard Part 2 §Header Attributes.

## Backward compatibility

The fields are added to a public interface (`IMTConnectStreamsHeader`). Any third-party consumer that implements this interface manually (rather than going through `MTConnectStreamsHeader`) needs to add the two new properties — but that's a small surface in practice. The default values (`0`) match the spec's `use="optional"` semantic.

### DIFF
--- a/libraries/MTConnect.NET-Common/Agents/MTConnectAgentBroker.cs
+++ b/libraries/MTConnect.NET-Common/Agents/MTConnectAgentBroker.cs
@@ -486,6 +486,8 @@ namespace MTConnect.Agents
             var header = new MTConnectStreamsHeader
             {
                 BufferSize = _observationBuffer.BufferSize,
+                AssetBufferSize = _assetBuffer != null ? _assetBuffer.BufferSize : 0,
+                AssetCount = _assetBuffer != null ? _assetBuffer.AssetCount : 0,
                 CreationTime = DateTime.UtcNow,
                 DeviceModelChangeTime = DeviceModelChangeTime.ToString("o"),
                 InstanceId = InstanceId,
--- a/libraries/MTConnect.NET-Common/Headers/IMTConnectStreamsHeader.cs
+++ b/libraries/MTConnect.NET-Common/Headers/IMTConnectStreamsHeader.cs
@@ -33,6 +33,16 @@ namespace MTConnect.Headers
         ulong BufferSize { get; }
 
         /// <summary>
+        /// A value representing the maximum number of Asset Documents that MAY be retained in the Agent that published the Response Document at any point in time.
+        /// </summary>
+        ulong AssetBufferSize { get; }
+
+        /// <summary>
+        /// A number representing the current number of Asset Documents that are currently stored in the Agent that published the Response Document.
+        /// </summary>
+        ulong AssetCount { get; }
+
+        /// <summary>
         /// A number representing the sequence number assigned to the oldest piece of Streaming Data stored
         /// in the buffer of the Agent immediately prior to the time that the Agent published the Response Document.
         /// </summary>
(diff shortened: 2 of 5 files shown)


ottobolyos:
Closing — based on a cross-source audit (`extra-files.user/upstream/issues/03-…`) the dime team's writeup mis-identified the surface.

Per `MTConnectStreams_2.7.xsd`, the Streams Header does NOT carry `assetBufferSize` or `assetCount`. Those fields are on `MTConnectDevices_2.7.xsd`'s Header and `MTConnectAssets_2.7.xsd`'s Header — which `IMTConnectDevicesHeader` and `IMTConnectAssestsHeader` already correctly expose. cppagent's `streamHeader()` likewise skips them and only `probeAssetHeader()` emits them.

This PR added them to `IMTConnectStreamsHeader`, which would have been a regression against the spec. Closing without merge. The two existing Header surfaces remain correct.

PR #169, opened 2026-05-25
fix(common): InvalidDeviceAdded event + universal ValidationResult

ottobolyos:
## Summary

Closes the startup-validation half of the device-identity contract that issue #136 / #137 already addressed at the constructor level, using the event-based pattern the rest of the `MTConnectAgent` `Invalid*Added` family already uses.

The parameterless `Device()` constructor (and `Agent()` by inheritance) was cleaned up in commit `38fa328e` so it leaves `Id`, `Name`, and `Uuid` all `null` -- the operator's config is now the sole source of identity. That left a follow-on gap: if a caller forgets to set `Uuid` before calling `MTConnectAgent.AddDevice`, the omission previously surfaced as an opaque `ConcurrentDictionary` `ArgumentNullException` far from the config gap that caused it.

This PR adds the missing event surface and consolidates the parallel `ValidationResult` structs that the rest of the family already uses.

### Validation surface

- New `MTConnect.ValidationResult` readonly struct (`libraries/MTConnect.NET-Common/ValidationResult.cs`) with `IsValid`, `Code` (machine-readable discriminator), `Message`, plus static `Valid()` / `Invalid(code, message)` factories.
- The three legacy variants are deleted (hard v7 move per maintainer):
  - `MTConnect.Devices.DataItems.ValidationResult` -- ~55 callsites migrated.
  - `MTConnect.Observations.ObservationValidationResult` -- was dead code; the only `Validate()` override of `IObservation` actually returned the DataItem variant.
  - `MTConnect.Assets.AssetValidationResult` -- migrated alongside `Asset.IsValid`, `FileAsset`, `CuttingToolAsset`, `IAsset`, `MTConnectAssetValidationHandler`, and `MTConnectAgentApplication.InvalidAsset`.

### Agent surface

- New `InvalidDeviceAdded` event on `MTConnectAgent` (and declaration on `IMTConnectAgent`), matching the existing `InvalidComponentAdded` / `InvalidCompositionAdded` / `InvalidDataItemAdded` / `InvalidObservationAdded` / `InvalidAssetAdded` family verbatim.
- New `MTConnectAgent.ValidateDevice(IDevice device)` helper so callers can pre-flight a device without subscribing.
- `AddDevice` no longer throws `InvalidOperationException`; it now raises `InvalidDeviceAdded` with a `ValidationResult` carrying `Code` (`"DeviceUuidMissing"` or `"DeviceNull"`) and a `Message`, then returns.

Validation codes introduced in this PR: `"DeviceUuidMissing"`, `"DeviceNull"`. A central code registry (enum or constants class) is left as a follow-up if the maintainer wants one.

## Test plan

- `DeviceCtorStartupValidationTests`: 4 converted (throw -> event) + 3 new (`ValidateDevice` pre-flight, null `Device` negative, happy-path no-event positive). 7 total, all green.
- Full default-gate suite: 4 956 / 4 956 passed across 8 assemblies.
- `RequiresDocker`: 4 / 4 passed.
- `E2E`: 4 / 4 passed.
- `XsdLoadStrict`: 36 / 36 passed.
- Integration coverage run (`-p:IntegrationCoverage=true`): 16 / 16 passed.
- Full solution Debug build: 0 errors. Warning delta vs `master`: zero new categories (1068 vs 1062 from duplicated multi-TFM warnings).

## Spec citation

- `MTConnectDevices_2.7.xsd` line 5245: `<xs:attribute name='uuid' type='UuidType' use='required'>`
- `MTConnectDevices_2.7.xsd` lines 147 to 155: `UuidType` documented as "uniquely identifies the element for it's entire life"

## Conventions

No `Co-Authored-By` trailer per project conventions; commits are ottobolyos-only.


PatrickRitchie:
I agree this is needed to prevent the Device from being added but I would prefer if it didn't throw an exception. I would think raising an Event would be better as it can still be logged but would prevent the entire app from potentially stopping.

The AddDevice() method may be called during runtime depending on the implementation and could cause the entire app to stop. I think the MTConnectAgent class should ideally be able to run without worrying about it throwing an unhandled exception and should only stop when explicitly told to via the Stop() method.

Let me know your thoughts.

ottobolyos:
The throw was overkill—I agree. I'll switch `AddDevice` to follow the existing `Invalid*Added` event family on `MTConnectAgent` (the pattern you introduced in 2022 — `InvalidComponentAdded`, `InvalidCompositionAdded`, `InvalidDataItemAdded`, `InvalidObservationAdded`, `InvalidAssetAdded`): a new `InvalidDeviceAdded` event raises with the device + reason, the device is not added, and the method returns without throwing. That keeps the agent running and gives downstream consumers a single, consistent surface to log against—they're likely already wiring the rest of the family.

Two design notes I'd appreciate your call on:

- **Event name.** `InvalidDeviceAdded` parallels the existing family, but it's a slight tense oddity since the device isn't actually added. Would you prefer `DeviceAddRejected` (more accurate) or stick with the family-consistent `InvalidDeviceAdded`?
- **Reason payload.** The previous throw distinguished null-Uuid from empty-Uuid in its message. I can collapse to a single `"Uuid is required"` string, or pass a `reason` parameter on the handler so consumers can switch on it—I'll go with whichever you prefer.

I'll rework the four existing tests in the same commit to assert (event-raised + device-not-added) instead of (exception-thrown).

PatrickRitchie:
- Event Name : Yes `InvalidDeviceAdded` is the preferred name in order to fit with the existing events. "Rejected" is probably the proper term but isn't consistent with what I had already named the others.

- Reason payload : I like the detailed message with a reason. I pass a ValidationResult struct in the existing events that should probably have a "reason" or "code" property to try and standardize the error along with a more detailed 'message' property. You can use the same ValidationResult struct or create something new. If you use the ValidationResult, we may want to change the namespace as it currently lives in the MTConnect.Devices.DataItems and should probably be changed to MTConnect.Devices or just under MTConnect as it is used for more than just DataItems. There is also an ObservationValidationResult struct with the same properties that could be replaced with a more universal struct. This will all be implemented in a new major version (v7) so I would be ok with changes similar to this.

PatrickRitchie (approved):
Looks good!

PR #177, opened 2026-05-30
fix: Condition observations not relayed when durableRelay is enabled

Pranay265:
When durableRelay: true, the Condition category observations were not getting relayed. the durable condition code path passes raw IObservationOutput objects directly to MTConnectMqttEntityServer.CreateMessage, which requires observations.DataItem.Device to be set. These raw outputs have Device == null , so CreateMessage() returns null →  throws → silently swallowed.

### DIFF
--- a/agent/Modules/MTConnect.NET-AgentModule-MqttRelay/Module.cs
+++ b/agent/Modules/MTConnect.NET-AgentModule-MqttRelay/Module.cs
@@ -899,28 +899,19 @@ namespace MTConnect
                                 if (observation.Category == DataItemCategory.CONDITION)
                                 {
                                     var conditionObservations = Agent.GetCurrentObservations(observation.DeviceUuid, observation.DataItemId);
-                                    var wrapped = conditionObservations?.Select(obs =>
-                                            {
-                                                var x = new Observation();
-                                                x.DeviceUuid = obs.DeviceUuid;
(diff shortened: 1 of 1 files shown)


Pranay265:
Hi @PatrickRitchie 

Just following up on this PR. The issue fixed here is critical for our application, and we would greatly appreciate a review when possible. Please let me know if there are any concerns or changes required from our side.

Thanks!

PatrickRitchie (approved):
Sorry for the delay in reviewing. There were some changes made to that same method that I needed to pull in first.

I did change the .Select(o => ...) LINQ statement to use the CloneAsObservation() method that was added in a previous commit. This should perform the same way and will make the code easier to maintain.

Thanks for creating a PR and please feel free to create others!

Pranay265:
Thanks for the update and for reviewing the PR.

Could you let us know when the next release is expected? We need to update our application once a version containing this change is available.

Pranay265:
Hi @PatrickRitchie 

Thanks again for reviewing and merging the fix in #177. I saw that it's included in the latest pre-release, thanks for getting that published.

We're planning to upgrade our production deployment once a stable release includes this fix, so I wanted to ask if you have a rough estimate for when the next stable release might be available.

Also, if the current pre-release is considered production-ready, I'd be interested to hear your recommendation.

PR #193, opened 2026-06-11
fix(common): coerce null/empty/whitespace observation Result to UNAVAILABLE

ottobolyos:
## Summary

`MTConnectAgent.AddObservation` accepted observations whose Result value was `null`, the empty string `""`, or whitespace-only and forwarded that value verbatim onto every wire transport (HTTP `/current`, `/sample`, MQTT JSON-cppagent, SHDR). MTConnect Part 1 §Observation Information Model — Representation — Observation Values mandates `"UNAVAILABLE"` as the sole valid representation of a missing or undetermined observation value:

> "If an Agent cannot determine a Valid Data Value for a DataItem, the value returned for the Result for the Data Entity MUST be reported as UNAVAILABLE."

The empty string is, by definition, not a Valid Data Value (it parses as no value at all under every typed DataItem schema). This PR adds an unconditional coerce step in the canonical `AddObservation(string deviceKey, IObservationInput, …)` overload — null / empty / whitespace Result is rewritten to `Observation.Unavailable` before validation, so:

- **`Ignore` / `Warning` / `Remove`**: publishes `UNAVAILABLE` rather than the empty value, satisfying the spec mandate on the wire.
- **`Strict`**: coerces to `UNAVAILABLE`, the observation lands rather than being silently dropped.

The coerce skips CONDITION observations (their Level enum cannot carry the empty pathology). DATA_SET / TABLE / TIME_SERIES Representation pre-fill of Count / SampleCount runs after the coerce — semantics for those representations are unchanged.

No opt-out flag is exposed: the spec mandate is `MUST`, so a way to opt out of compliance would itself be non-conformant.

## Behaviour change

| Input | Pre-fix wire payload | Post-fix wire payload |
|---|---|---|
| Result = `null` | `""` (empty) | `"UNAVAILABLE"` |
| Result = `""` | `""` | `"UNAVAILABLE"` |
| Result = `"   "` (whitespace) | `"   "` | `"UNAVAILABLE"` |
| Result = `"AVAILABLE"` (concrete) | `"AVAILABLE"` | `"AVAILABLE"` (unchanged) |
| Result = `""` under `Strict` | observation silently dropped | lands with `"UNAVAILABLE"` |

## Tests

**Unit** (NUnit, `tests/MTConnect.NET-Common-Tests/Agents/AddObservationEmptyResultCoerceTests.cs`, 10 cases):

- empty-string Result under every non-Strict `InputValidationLevel`
- the `null` / `""` / whitespace family under `Warning`
- empty Result under `Strict` (lands, not silently dropped)
- concrete-value preservation (negative — the coerce does not substitute for a Valid Data Value)

**Wire-level E2E** (xUnit, `[Trait("Category", "E2E")]`):

`tests/MTConnect.NET-Integration-Tests/Workflows/AddObservationEmptyResultUnavailableWorkflowTests.cs` — HTTP `/current`:
- null / empty / spaces / tab / newline / CRLF Result → AVAILABILITY element body is `UNAVAILABLE` (6-case `[Theory]`)
- concrete value → AVAILABILITY element body is verbatim
- concrete → empty sequence → AVAILABILITY element reflects the latest (`UNAVAILABLE`)
- empty Result under `Strict` validation → AVAILABILITY element body is `UNAVAILABLE` (lands, not silently dropped; spins its own agent + server harness)

`tests/MTConnect.NET-Integration-Tests/Workflows/AddObservationEmptyResultUnavailableSampleStreamWorkflowTests.cs` — HTTP `/sample` stream:
- null / empty / spaces / tab / newline / CRLF Result → latest AVAILABILITY sample is `UNAVAILABLE` (6-case `[Theory]`)
- concrete value → latest AVAILABILITY sample is verbatim
- concrete → empty sequence → two distinct samples in stream order (concrete first, then `UNAVAILABLE`); proves the coerce does not silently drop the second observation

All wire surfaces fed by the agent's buffer read the same coerced payload — every transport (HTTP, MQTT relay, SHDR adapter output) inherits the fix because the coerce sits above the buffer.

## Files touched

- `libraries/MTConnect.NET-Common/Agents/MTConnectAgent.cs` — invoke the coerce in `AddObservation`; add `IsEmptyResult` + `CoerceEmptyResultToUnavailable` helpers.
- `tests/MTConnect.NET-Common-Tests/Agents/AddObservationEmptyResultCoerceTests.cs` — new unit fixture.
- `tests/MTConnect.NET-Integration-Tests/Workflows/AddObservationEmptyResultUnavailableWorkflowTests.cs` — new wire-level E2E fixture against HTTP `/current`.
- `tests/MTConnect.NET-Integration-Tests/Workflows/AddObservationEmptyResultUnavailableSampleStreamWorkflowTests.cs` — new wire-level E2E fixture against HTTP `/sample` stream.
- `.github/workflows/dotnet.yml` — drop matrix expressions from the two matrix jobs' `name:` fields so skipped-state status checks render cleanly (`build-and-test` / `route-check-e2e` instead of `build-and-test-${{ matrix.os }}`).
- `docs/reference/configuration.md` — regenerated to surface the new `deviceValidationLevel` config key on the docs site (`IAgentConfiguration` + `AgentConfiguration` sections). Same regenerator (`docs/scripts/generate-reference.sh`) that drives the docs-site drift gate.

## Additional scope (from cascade rebase)

Includes maintainer commit `a61a3647 Added separate DeviceValidationLevel propery and enum to handle MTConnectDevices validation`:

- **New enum `MTConnect.Agents.DeviceValidationLevel`** (`Ignore` / `Warning` / `Remove` / `Strict`) — controls how the agent reacts when `MTConnectDevices` shape data fails validation against the standard.
- **New `AgentConfiguration.DeviceValidationLevel` / `IAgentConfiguration.DeviceValidationLevel`** property (default `Warning`). Serialised key: `deviceValidationLevel`.
- **`MTConnectAgent.NormalizeDevice`** — the three invalid-Component / -Composition / -DataItem branches now consult `DeviceValidationLevel` instead of `InputValidationLevel`, splitting device-shape validation from observation validation. Enables `InputValidationLevel = Strict` (spec-conformant observations) with `DeviceValidationLevel = Warning` (accept non-standard device shapes).

Orthogonal to the empty-Result coerce contract above; the coerce runs unconditionally in `AddObservation` regardless of either level, per the Part 1 `MUST`.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -36,7 +36,7 @@ jobs:
   # paid exactly once across the whole workflow (in job 2 below).
   # ------------------------------------------------------------------
   build-and-test:
-    name: build-and-test-${{ matrix.os }}
+    name: build-and-test
     # Skip drafts: run only on push-to-master + ready (non-draft) PRs.
     # The pull_request `types` list above includes `ready_for_review`
     # so CI fires the moment a draft is flipped to ready.
@@ -323,7 +323,7 @@ jobs:
   # categories run exactly once (in job 1) across the workflow.
   # ------------------------------------------------------------------
   route-check-e2e:
-    name: route-check-e2e-shard${{ matrix.shard }}of${{ matrix.shardTotal }}
+    name: route-check-e2e
     if: github.event_name == 'push' || github.event.pull_request.draft == false
     needs: docs-prepare
     runs-on: ubuntu-latest
--- a/libraries/MTConnect.NET-Common/Agents/MTConnectAgent.cs
+++ b/libraries/MTConnect.NET-Common/Agents/MTConnectAgent.cs
@@ -2152,6 +2152,17 @@ namespace MTConnect.Agents
                 var dataItem = GetDataItem(deviceUuid, input.DataItemKey);
                 if (dataItem != null)
                 {
+                    // Coerce a null, empty, or whitespace-only Result to UNAVAILABLE before the
+                    // observation reaches the buffer. The MTConnect Part 1 Observation Information
+                    // Model mandates "UNAVAILABLE" as the sole valid representation of a missing
+                    // value; an empty string is never a Valid Data Value. The coerce runs above
+                    // validation so Strict no longer silently drops empty-Result observations and
+                    // every other validation level publishes the spec-compliant sentinel.
+                    if (dataItem.Category != DataItemCategory.CONDITION && IsEmptyResult(input))
+                    {
+                        CoerceEmptyResultToUnavailable(input);
+                    }
+
                     // Add required properties
                     switch (dataItem.Representation)
                     {
@@ -2274,6 +2285,41 @@ namespace MTConnect.Agents
             }
 
             return false;
+        }
+
+
+        /// <summary>
+        /// Returns true when the observation's Result value is null, the empty string, or whitespace-only.
+        /// </summary>
+        /// <remarks>
+        /// An empty Result is, by definition, not a Valid Data Value under any typed DataItem schema; the
+        /// MTConnect Part 1 Observation Information Model mandates "UNAVAILABLE" as the sole valid
+        /// representation of a missing value. This predicate identifies the inputs the SDK coerces.
+        /// </remarks>
+        private static bool IsEmptyResult(IObservationInput input)
+        {
+            var result = input.GetValue(ValueKeys.Result);
+            if (result == null) return true;
+            return string.IsNullOrWhiteSpace(result);
+        }
+
+        /// <summary>
+        /// Rewrites the observation's Result value to <see cref="Observation.Unavailable"/> and flags the input as unavailable.
+        /// </summary>
+        /// <remarks>
+        /// Removes any prior Result entry (so the Values collection does not carry a duplicate ValueKey),
+        /// adds the UNAVAILABLE sentinel, and sets <see cref="IObservationInput.IsUnavailable"/> so downstream
+        /// consumers that branch on the flag observe the coerced state. Spec authority: MTConnect Part 1
+        /// Observation Information Model - Representation - Observation Values.
+        /// </remarks>
+        private static void CoerceEmptyResultToUnavailable(IObservationInput input)
+        {
+            var preserved = (input.Values ?? Enumerable.Empty<ObservationValue>())
+                .Where(v => v.Key != ValueKeys.Result)
+                .ToList();
+            input.Values = preserved;
+            input.AddValue(ValueKeys.Result, Observation.Unavailable);
+            input.IsUnavailable = true;
         }
 
 
--- a/tests/MTConnect.NET-Common-Tests/Agents/AddObservationEmptyResultCoerceTests.cs
+++ b/tests/MTConnect.NET-Common-Tests/Agents/AddObservationEmptyResultCoerceTests.cs
@@ -1,0 +1,142 @@
+// Copyright (c) 2026 TrakHound Inc., All Rights Reserved.
+// TrakHound Inc. licenses this file to you under the MIT license.
+
+using System;
+using System.Linq;
+using MTConnect.Agents;
+using MTConnect.Configurations;
+using MTConnect.Devices;
+using MTConnect.Devices.DataItems;
+using MTConnect.Observations;
+using NUnit.Framework;
+
+namespace MTConnect.Tests.Common.Agents
+{
+    /// <summary>
+    /// Regression pin for the empty-Result wire-level spec violation in
+    /// <see cref="MTConnectAgent.AddObservation(string, MTConnect.Input.IObservationInput, bool?, bool?, bool?, bool)"/>:
+    /// when the inbound observation carries a null, empty, or whitespace-only
+    /// Result value, the agent currently publishes that value verbatim on the
+    /// wire. The MTConnect Part 1 Observation Information Model mandates
+    /// "UNAVAILABLE" as the sole valid representation of a missing value, so
+    /// the SDK must coerce null / empty / whitespace to
+    /// <see cref="Observation.Unavailable"/> before the observation reaches
+    /// the buffer.
+    ///
+    /// Spec: MTConnect Standard, Part 1 - Devices Information Model,
+    /// Observation Information Model - Representation - Observation Values
+    /// ("If an Agent cannot determine a Valid Data Value for a DataItem, the
+    /// value returned for the Result for the Data Entity MUST be reported as
+    /// UNAVAILABLE.").
+    ///
+    /// This fixture pins the coerce on the canonical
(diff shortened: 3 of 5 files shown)


ottobolyos:
@PatrickRitchie, this PR fixes a Part 1 spec violation that touches every wire output: `MTConnectAgent.AddObservation` was forwarding null / empty / whitespace Result values verbatim, where Part 1 §Observation Information Model — Representation — Observation Values mandates `UNAVAILABLE` as the sole valid representation of a missing value. The fix is small and self-contained—4 files across 4 commits, with 10 unit + 17 wire-level E2E cases pinning the contract—but the bug class is broad enough (every HTTP envelope, MQTT relay, and SHDR adapter output benefits from the coerce at the buffer entry) that a review at your convenience would be appreciated.

ottobolyos:
@PatrickRitchie — heads-up: while validating PR #189's dime-connector build I caught an unrelated multi-TFM Release-pack regression introduced by an earlier warnings sweep. The minimal fix lands in PR #194 (just opened as draft); the broader warnings cleanup plus a Release-pack CI gate ships in PR #195.

PatrickRitchie:
## Updates
I updated the PR to allow this feature to be toggled on/off based on the 'InputValidationLevel = strict' configuration parameter. This should allow any existing implementation to still work while allowing a strict adherence to the standard when required.

I also added a separate DeviceValidationLevel configuration parameter in order to allow a 'non-standard' DataItem or Component to work with the 'strict' InputValidationLevel. This is common for end users to use a DataItem with a custom Type or SubType.

## Tests
I'm not sure extactly how this affects the tests but I saw where InputValidationLevel is used a few times so that may be why it was failing.

ottobolyos:
@PatrickRitchie, following up on the empty-Result coerce in `MTConnectAgent.AddObservation` (line 2161 of `libraries/MTConnect.NET-Common/Agents/MTConnectAgent.cs`) — commit `8dd28c6b Updated to toggle on/off based on input validation level set to 'strict'` narrowed the coerce guard to fire only when `InputValidationLevel == Strict`, and that introduces a real MTConnect standard violation on the other three levels.

Under `Ignore`, `Warning`, and `Remove` (with `Warning` being the default), a null / empty-string / whitespace-only Result now flows verbatim onto every wire transport — HTTP `/current` and `/sample`, MQTT JSON-cppagent, SHDR.

The Part 1 mandate this repo already recognises verbatim (see `libraries/MTConnect.NET-Common/Observations/Observation.cs:20–27`):

> If an Agent cannot determine a Valid Data Value for a DataItem, the value returned for the Result for the Data Entity MUST be reported as UNAVAILABLE.

The empty string is by definition not a Valid Data Value under any typed DataItem schema:

- Every `MTConnectStreams_*.xsd` restriction that permits `UNAVAILABLE` does so as an *enumerated string token*, never as an empty content model — see e.g. `tests/Compliance/MTConnect-Compliance-Tests/Schemas/v2_2/MTConnectStreams_2.2.xsd:4567` and every enumerated Event/Sample restriction on that schema.
- On `xs:decimal`-typed Samples the empty string flat-out fails XSD validation (`""` is not a valid `xs:decimal` lexical form).
- On Events with typed enums the empty string is not among the enumerated tokens either.
- cppagent — the MTConnect reference implementation — coerces or rejects empty Results at ingress and never emits them on the wire.

The 8 red tests on `AddObservation_NullEmptyOrWhitespaceResult_Coerced_To_Unavailable` (null / `""` / `"   "` / `"\t"` / `"\n"` / `"\r\n"` / …) pin the pre-`8dd28c6b` behaviour and are what caught this. Under the current guard they all fail because the coerce is skipped.

## Proposal

Restore the unconditional coerce — drop the `&& _configuration.InputValidationLevel == InputValidationLevel.Strict` clause so the guard reads `dataItem.Category != DataItemCategory.CONDITION && IsEmptyResult(input)`. Every level then publishes the spec-mandated `UNAVAILABLE`; Strict still gains the original benefit (the empty observation lands rather than being silently dropped by the validator). All 8 tests stay green as they are.

If you want the pre-coerce raw-empty behaviour preserved as an opt-in — e.g. for a downstream tool that inspects the raw payload — we can add an `AgentConfiguration.CoerceEmptyResultToUnavailable` boolean defaulting to `true`. Flipping it to `false` produces an agent that is not fully conformant with MTConnect Part 1, so we'd log a startup warning describing the non-conformance and gate the flip behind that warning, but the mechanism is there for downstream consumers who genuinely need it.

Happy to push the unconditional-coerce fix directly. #193 stays draft until we agree on the shape.

ottobolyos:
@PatrickRitchie, refreshing the merge-train snapshot now that #205 has landed and everything downstream has cascade-rebased onto `587a1265`. Seven drafts remain, ordered per CONVENTIONS §1.5b:

1. **#193** `fix/agent-empty-result-unavailable`—coerce empty `Result` → `UNAVAILABLE` per Part 1's MUST;
2. **#194** `fix/multi-tfm-compat`—`SupportedOSPlatform` NET5+ guard + XML `LangVersion` 8.0;
3. **#195** `fix/release-pack-warnings-cleanup`—CA2022 short-read + SYSLIB0057 + net4x lifetime restore (carries the pre-cascade APPROVE forward per §1.5d);
4. **#207** `fix/agent-uuid-lifecycle-hardening`—`AgentUuidResolver`; RFC 4122 validation on both the operator override and the persisted `agent.information.json` paths; sanitised log echo;
5. **#208** `fix/jsonconditions-structural-rewrite`—deletes `JsonConditionsConverter` (~200 lines), rewrites `JsonConditions` as `List<JsonConditionWrapper>`-derived so `System.Text.Json` handles both directions with no custom code;
6. **#213** `fix/agent-config-sender-writable`—`IAgentConfiguration.Sender { get; }` + concrete `{ get; set; }` + ctor wiring, so `<Header sender="…">` per Part 1 §7 is authorable via configuration rather than pinned to `Dns.GetHostName()`;
7. **#214** `feat/configuration-vendor-extensions`—`IConfiguration.VendorExtensions` (`IEnumerable<XElement>`) delivered inside the `<Configuration>` sequence via the v2.7 XSD's `AbstractConfiguration` substitutionGroup mechanism, so a vendor XSD whose element substitutes `AbstractConfiguration` rides the standard's own extensibility hook rather than a raw-XML side-channel.

Downstream sanity: `integration/all-fixes` at `9acc579d` unions the original five on top of `587a1265`; #213 and #214 will cascade in on the next refresh. A full multi-TFM pack round-trips through `dime-connector` locally. Full write-ups sit in each PR body; I'm happy to answer anything inline.

The one open thread with the ball still on your side is the coerce-contract question I left inline on #193 on 16 July 2026—whether Part 1's MUST reads as an unconditional coerce or as a policy knob. The wire fix is decoupled from that call and doesn't block on it; the question shapes how the follow-up docs are worded, not the code.

